Issue #11, opened 2023-03-25
Allow screenshots to be resized


takahirom:
This would be an essential feature, as users would not be able to change the size without this feature when using the velify task.
I am considering a parameter that allows the user to pass a value of 0~1, so that passing 0.5 would result in 50%.

resize = 0.5

PR #13, opened 2023-03-25
Screenshot image size can now be adjusted

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/11


takahirom:
Comparison images were not generated and the cause is under investigation.

takahirom:
It seems working.
image

Issue #33, opened 2023-04-19
captureRoboImage() of a View and Composable, without the need of ViewInteractions or SemanticNodes


sergio-sastre:
### The issue
Currently, Roborazzi depends heavily on Espresso and ComposeRule.

That's not the case with other screenshot testing frameworks, like Paparazzi, Shot, Dropshots...

In screenshot testing, it's also common to inflate a View without attaching it to an Activity. **If not attached, you cannot find it with Espresso, and therefore it would not be possible to screenshot it**.

Moreover, not attaching a View to an Activity has one extra advantage:
you can control its width and height to see how the View would render for different widths and heights.

That's something AndroidUiTestingUtils provides via `waitForMeasureView(exactWidthPx, exactHeightPx)`, like in [this example here](https://github.com/sergio-sastre/Android-screenshot-testing-playground/blob/master/dialogs/shot/src/androidTest/java/com/example/road/to/effective/snapshot/testing/dialogs/shot/parameterized/DeleteDialogWithParameterizedRunnerTest.kt), but it's not compatible with Roborazzi because of the reason I've described above.

### Desired Solution
Ideally, `captureRoboImage(view)` and `captureRoboImage(@Composable () -> Unit)` should be provided to enable such cases and become less Espresso/ComposeRule dependent.

Additionally, a similar option for bitmaps e.g. `captureRoboImage(bitmap)` would be also useful, since you can draw a bitmap out of a View/Composable but applying your own Bitmap options. Shot and Dropshots offer that as well.

takahirom:
Thanks.
I think it is possible for Android View, probably because the View probably has a context and theme.
In the case of Compose, I think it is possible by doing ActivitySenario.launch, etc.
My question is in the case of View, how do you instantiate the View in your test?

sergio-sastre:
That's the trick of `ActivityScenarioConfigurator.ForView()` and `ActivityScenarioConfigurator.ForComposable()`: They launch an activity with a specific configuration for locale, fontscale, uiMode... and use the context of that activity to inflate views -> the views inherit the configuration of that context ;)

I reommend you take the sample I mentioned previously:
[Dialog screenshot example](https://github.com/sergio-sastre/Android-screenshot-testing-playground/blob/master/dialogs/shot/src/androidTest/java/com/example/road/to/effective/snapshot/testing/dialogs/shot/parameterized/DeleteDialogWithParameterizedRunnerTest.kt)

[...]

PR #42, opened 2023-04-21
Support View, Compose lambda, Bitmap

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/33

image

### DIFF
--- a/app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
+++ b/app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
@@ -1,7 +1,10 @@
 package com.github.takahirom.roborazzi.sample
 
+import android.app.AlertDialog
 import android.graphics.Bitmap
 import android.view.View
+import android.view.ViewGroup
+import android.widget.TextView
 import androidx.compose.material3.Text
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithTag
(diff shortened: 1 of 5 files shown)


app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
```diff
@@ -58,6 +62,21 @@ class ManualTest {
 
     onView(ViewMatchers.isRoot())
       .captureRoboImage()
+
+    composeTestRule.activity.findViewById<View>(R.id.button_second)
+      .captureRoboImage("$DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH/manual_small_view_button_second_by_view.png")
```
takahirom, line 67: image

app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
```diff
@@ -58,6 +62,21 @@ class ManualTest {
...
+
+    captureRoboImage {
+      Text("Hello Compose!")
```
takahirom, line 70: image

app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
```diff
@@ -58,6 +62,21 @@ class ManualTest {
...
+    }
+
+    createBitmap(100, 100, Bitmap.Config.ARGB_8888)
+      .apply {
+        applyCanvas {
+          drawColor(android.graphics.Color.RED)
+        }
+      }
+      .captureRoboImage("$DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH/manual_small_view_button_second_by_bitmap.png")
```
takahirom, line 79: image

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -72,6 +82,89 @@ fun ViewInteraction.captureRoboImage(
...
+  roborazziOptions: RoborazziOptions = RoborazziOptions(),
+) {
+  if (!roborazziEnabled()) return
+  captureRoboImage(
+    file = File(filePath),
+    roborazziOptions = roborazziOptions
+  )
+}
+
+fun View.captureRoboImage(
```
sergio-sastre, line 96: I've tested this and unfortunately, it looks like I was right: it would not work for a Dialog since the view is not attached to the activity window.

However, since Bitmaps would be supported too, one could call `dialog.window.decorView.drawToBitmap().captureRoboImage()` and should be fine I believe... (could not test it)

Another option would be to get it via Espresso... but I could not find a way...

I'd prefer a solution without Espresso, but could serve the purpose for a first iteration?

sergio-sastre (approved):

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
takahirom: I attempted and implemented a solution, but I couldn't resolve an issue that arises when displaying a dialog stating, "java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first." This occurs even when I remove the view from the parent.

https://github.com/takahirom/roborazzi/pull/42/commits/5b7e51851ce514848a80143929d000f5cb60fa4a#diff-1bb3d1a101c61df64c19694d660a430f6802d62054f0963f804d9edca69fb971R86

Regardless, I believe it might be better than having no implementation, so I would like to introduce them.

app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
```diff
@@ -58,6 +63,26 @@ class ManualTest {
...
     onView(ViewMatchers.isRoot())
       .captureRoboImage()
+
+    composeTestRule.activity.findViewById<View>(R.id.button_second)
+      .captureRoboImage("$DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH/manual_small_view_button_second_by_view.png")
+
+    TextView(composeTestRule.activity).apply {
+      text = "Hello View!"
+      setTextColor(android.graphics.Color.RED)
+    }.captureRoboImage("$DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH/manual_view_without_window.png")
```
takahirom, line 73: image

Issue #34, opened 2023-04-19
Resulting RoboImage has black background, although ActivityScenario background is set to TRANSPARENT


sergio-sastre:
Since Robolectric cannot render shadows, it is pretty useful to provide a background color to the activity launched via ActivityScenario to better distinguish the "borders", e.g. in CardViews.

AndroidUiTestingUtils offers that possibility like this (same for composables), for instance:
```kotlin
@get:Rule
val activityScenarioForViewRule =
        ActivityScenarioForViewRule(
            config = ViewConfigItem(...),
            backgroundColor = Color.TRANSPARENT,
        )
```

When used with Roborazzi, if backgroundColor = Color.TRANSPARENT, the background color is eventually black (I believe, it is might be set to null?). It works with other colours though

Moreover, I believe it is related to Roborazzi and not Robolectric, since if I use the following method `compareSnapshot`, instead of `captureRoboImage()`, the background is actually transparent...

```kotlin
fun compareSnapshot(view: View, name: String) {
    val image = view.drawToBitmap()
    val path = System.getProperty("user.dir")
    val file = File("$path/src/test", "$name.png")

    try {
        val bos = ByteArrayOutputStream()
        val out = FileOutputStream(file)
        out.write(bos.toByteArray());
        image.compress(Bitmap.CompressFormat.PNG, 100, out)
        out.flush()
        out.close()
    } catch (e: Exception) {
        e.printStackTrace()
    }
}
```

### Current result with TRANSPARENT backgroundColor:
With `captureRoboImage()`
![MemoriseViewHolder_Happy_full](https://user-images.githubusercontent.com/6097181/233208128-dd3a896b-63c6-427e-b43f-a12cf7f3a737.png)

With `compareSnapshot()`:
![MemoriseViewHolder_Happy_compare](https://user-images.githubusercontent.com/6097181/233207929-3426c111-e2a4-47a1-95f6-cb65e11316f6.png)

takahirom:
Thank you for reporting this issue.
It could be due to this property that the library was designed for dumping layouts. As a result, this property offers good performance. However, I believe it is not suitable for taking screenshots.

https://github.com/takahirom/roborazzi/blob/409dc0026e416c8f9418114c47c5f3720d498ef0/roborazzi-painter/src/main/java/com/github/takahirom/roborazzi/RoboCanvas.kt#L21

sergio-sastre:
[...]

PR #45, opened 2023-04-22
Support Wear screenshots

yschimke:
Not sure about the right API, particularly to support both full screen screenshots, and component screenshots without a crop.

image

### DIFF
--- a/app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeTest.kt
+++ b/app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeTest.kt
@@ -1,8 +1,17 @@
 package com.github.takahirom.roborazzi.sample
 
 import androidx.activity.ComponentActivity
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.Text
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.onRoot
 import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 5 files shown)


takahirom:
👀

yschimke:
I'm struggling to get a transparent background, so still working on it.

yschimke:
Examples of current RNG screenshots. https://github.com/joreilly/Confetti/blob/main/wearApp/snapshot/SessionsScreenTest/sessionsScreen_Colors_MobileAndroid__reference.png

takahirom:
Thanks, I think this issue might be the cause. 👀 
https://github.com/takahirom/roborazzi/issues/34

roborazzi/src/main/java/com/github/takahirom/roborazzi/capture.kt
```diff
@@ -285,7 +285,8 @@ data class RoborazziOptions(
   }
 
   data class RecordOptions(
-    val resizeScale: Double = 1.0
+    val resizeScale: Double = 1.0,
+    val fullScreen: Boolean = true
```
takahirom, line 289: I believe adding the parameter to RecordOptions is the right choice! However, the name "fullscreen" might be a bit confusing. Users could potentially think it refers to a navigation bar or status bar. I think the name "applyDeviceCrop: Boolean = true" is a better alternative and should be clearer for users.
While "fullscreen" is useful for other features and might be a good name, do you have any suggestions for features that would work well with fullscreen?

yschimke: Done.  I think eventually it should be true for full screen tests, and defaulted to true.  But for now, changing to false until it works end to end.

app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeTest.kt
```diff
@@ -33,4 +54,26 @@ class ComposeTest {
         .performClick()
     }
   }
+
+  @Test
+  @Config(
+    sdk = [30],
+    qualifiers = "w221dp-h221dp-small-notlong-round-watch-xhdpi-keyshidden-nonav"
```
takahirom, line 61: 👍

takahirom (approved):
I think it's better to include this implementation in the library. This pull request is marked as "For discussion." However, can I merge it?

app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeTest.kt
yschimke: We may want to provide these as "const val" for the same devices that are provided in Android Studio previews.

PR #48, opened 2023-04-23
Support image pixel type and make ARGB_8888 the default value.

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/34

Support image pixel type
~Currently, we are using RGB565 for performance reasons. However, transparency can be useful. For now, how about adding an option to use transparency, while leaving the default value as is?~

In this PR, I'm going to change the default image type to ARGB 8888. I'm considering making this the default setting, and I think this feature will be included in version 1.2. The reason is that we are getting more CPU and memory resources as time passes, and even in Android, Glide's default value is ARGB 8888.
https://bumptech.github.io/glide/doc/migrating.html#decodeformat

image
image


takahirom:
Since this library mistakenly used ARGB8888 for fetching view images, there are some breaking changes when verifying. Users should be notified of this issue.
image

yschimke:
The change makes sense to me

Issue #47, opened 2023-04-22
Add examples using robolectric @Config


yschimke:
Useful to show configuring the screen for different devices. Round vs Square watch.

These qualifiers should also affect which resources are loaded

```
@Config(
    sdk = [30],
    qualifiers = "w221dp-h221dp-small-notlong-notround-watch-xhdpi-keyshidden-nonav"
)
```

```
@Config(
    sdk = [30],
    qualifiers = "w221dp-h221dp-small-notlong-round-watch-xhdpi-keyshidden-nonav"
)
```

yschimke:
We could provide constants for known devices, similar to Paparazzi, and Android Studio.

```
@Preview(device = Devices.NEXUS_5X)
```

And config in https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:artwork/resources/device-art-resources/wearos_small_round/layout

https://github.com/cashapp/paparazzi/blob/c10c435bd30025f3b5da64f846c6ecbd66d54515/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt#L475

https://android.googlesource.com/platform/tools/base/+/mirror-goog-studio-master-dev/sdklib/src/main/java/com/android/sdklib/devices/wear.xml

sergio-sastre:
> We could provide constants for known devices, similar to Paparazzi, and Android Studio.
> 
> ```
> @Preview(device = Devices.NEXUS_5X)
> ```
> 
> And config in https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:artwork/resources/device-art-resources/wearos_small_round/layout
> 
> https://github.com/cashapp/paparazzi/blob/c10c435bd30025f3b5da64f846c6ecbd66d54515/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt#L475
> 
> https://android.googlesource.com/platform/tools/base/+/mirror-goog-studio-master-dev/sdklib/src/main/java/com/android/sdklib/devices/wear.xml

That is actually sth I’m releasing next week:
https://twitter.com/gio_sastre/status/1649073964617019400?s=46&t=ljy1XQW5SI3KUXpinEKEow

in this library:
https://github.com/sergio-sastre/AndroidUiTestingUtils

However, that is sth that would depend on ActivityScenario. So I believe, it might make sense to have this directly in Roborazzi?

[...]

PR #49, opened 2023-04-24
Add robolectric device qualifiers

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/47


roborazzi/src/main/java/com/github/takahirom/roborazzi/RobolectricDeviceQualifiers.kt
```diff
@@ -0,0 +1,153 @@
...
+    "w1374dp-h859dp-xlarge-long-notround-land-any-mdpi-keyshidden-nonav"
+  const val Device101WXGATabletPort =
+    "w1374dp-h859dp-xlarge-long-notround-port-any-mdpi-keyshidden-nonav"
+
+  // FROM: wear.xml
+  const val WearOSSquareLand = "w186dp-h186dp-small-long-round-land-watch-hdpi-keyshidden-nonav"
+  const val WearOSSquarePort = "w186dp-h186dp-small-long-round-port-watch-hdpi-keyshidden-nonav"
+  const val WearOSRoundLand = "w213dp-h213dp-small-long-round-land-watch-hdpi-keyshidden-nonav"
+  const val WearOSRoundPort = "w213dp-h213dp-small-long-round-port-watch-hdpi-keyshidden-nonav"
+  const val WearOSRoundChinLand =
```
yschimke, line 91: I don't think you need landscape and portrait here. Aspect is fixed on the wrist

roborazzi/src/main/java/com/github/takahirom/roborazzi/RobolectricDeviceQualifiers.kt
```diff
@@ -0,0 +1,153 @@
...
+    "w249dp-h226dp-small-notlong-notround-land-watch-tvdpi-keyshidden-nonav"
+  const val WearOSRoundChinPort =
+    "w249dp-h226dp-small-notlong-notround-port-watch-tvdpi-keyshidden-nonav"
+
+  // FROM: nexus.xml
+  const val NexusOnePort = "w302dp-h503dp-normal-long-notround-port-any-hdpi-keyshidden-trackball"
```
yschimke, line 97: A very personal style thingz I'd probably curate these somewhat, with just ones I expect to be used.

You might also find in practice, people apply the device, then switch portrait and landscape for a single test with "+land"

But all works for me as it includes wear.

yschimke (approved):

roborazzi/src/main/java/com/github/takahirom/roborazzi/RobolectricDeviceQualifiers.kt
```diff
@@ -0,0 +1,153 @@
...
+    "w7670dp-h4314dp-xlarge-long-notround-land-television-tvdpi-keyshidden-dpad"
+  const val AndroidTV720pPort =
+    "w7670dp-h4314dp-xlarge-long-notround-port-television-tvdpi-keyshidden-dpad"
+
+  // FROM: devices.xml
+  const val Device135FreeformLand =
+    "w1706dp-h960dp-xlarge-long-notround-land-any-hdpi-keyshidden-nonav"
+  const val Device135FreeformPort =
+    "w1706dp-h960dp-xlarge-long-notround-port-any-hdpi-keyshidden-nonav"
+  const val Device76FoldinwithouterdisplayPort =
```
sergio-sastre, line 21: My personal take: Device76 (or the like) tells me absolutely nothing.


I find it very confusing. The names should be device related.
Nobody knows what device135 is.

I'd rather do not try to be too clever downloading the devices from a file to generate this file but set them directly, like Paparazzi.

Otherwise you are risking to make breaking changes in the future, plus, you can easily make the names easy to read and understand, what is indeed the point of the feature imo

takahirom:
@sergio-sastre @yschimke 
I removed the "Port" and "Land" suffixes and filtered out unneeded devices. Could you take a look at this?
https://github.com/takahirom/roborazzi/pull/49/commits/a902dbd98b25ca16aa483d6700fb9226dbdfdc08

roborazzi/src/main/java/com/github/takahirom/roborazzi/RobolectricDeviceQualifiers.kt
```diff
@@ -4,150 +4,30 @@ object RobolectricDeviceQualifiers {
...
-  const val Pixel3aXLLand = "w434dp-h862dp-normal-long-notround-land-any-400dpi-keyshidden-nonav"
-  const val Pixel4Port = "w389dp-h821dp-normal-long-notround-port-any-440dpi-keyshidden-nonav"
-  const val Pixel4Land = "w389dp-h821dp-normal-long-notround-land-any-440dpi-keyshidden-nonav"
-  const val Pixel4XLPort = "w429dp-h905dp-normal-long-notround-port-any-560dpi-keyshidden-nonav"
-  const val Pixel4XLLand = "w429dp-h905dp-normal-long-notround-land-any-560dpi-keyshidden-nonav"
-  const val Pixel4aPort = "w390dp-h844dp-normal-long-notround-port-any-440dpi-keyshidden-nonav"
-  const val Pixel4aLand = "w390dp-h844dp-normal-long-notround-land-any-440dpi-keyshidden-nonav"
-  const val Pixel5Port = "w390dp-h844dp-normal-long-notround-port-any-440dpi-keyshidden-nonav"
-  const val Pixel5Land = "w390dp-h844dp-normal-long-notround-land-any-440dpi-keyshidden-nonav"
+  const val NexusOne = "w302dp-h503dp-normal-long-notround-any-hdpi-keyshidden-trackball"
```
takahirom, line 17: I would like to use this device for checking a small device 👀

app/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
```diff
@@ -31,15 +32,20 @@ import java.io.File
...
+@Config(
+  sdk = [30],
+  qualifiers = RobolectricDeviceQualifiers.NexusOne
+)
 class ManualTest {
   @get:Rule
   val composeTestRule = createAndroidComposeRule<MainActivity>()
 
-  @Test
+  @Test @Config(qualifiers = "+land")
```
yschimke, line 48: Nice!

yschimke (approved):
Looks great

roborazzi/src/main/java/com/github/takahirom/roborazzi/RobolectricDeviceQualifiers.kt
```diff
@@ -0,0 +1,33 @@
...
+  const val AndroidTV1080p = "w7670dp-h4314dp-xlarge-long-notround-television-xhdpi-keyshidden-dpad"
+  const val AndroidTV720p = "w7670dp-h4314dp-xlarge-long-notround-television-tvdpi-keyshidden-dpad"
+
+  // FROM: devices.xml
+// FROM: wear.xml
+  const val WearOSSquare = "w186dp-h186dp-small-long-round-watch-hdpi-keyshidden-nonav"
+  const val WearOSRound = "w213dp-h213dp-small-long-round-watch-hdpi-keyshidden-nonav"
+  const val WearOSRoundChin = "w249dp-h226dp-small-notlong-notround-watch-tvdpi-keyshidden-nonav"
+
+  // FROM: nexus.xml
```
sergio-sastre, line 16: @takahirom 
I've checked with the previous values, and for Pixel4a, height and weight were slightly different as on my device.

The qualifier was:
qualifiers = "w390dp-h844dp-normal-long-440dpi..."

But to make it look like in the phone: qualifiers = "w393dp-h847dp-normal-long-440dpi"

And I believe that is because you calculate width and height from the devices.xml file by using xdpi and ydpi.

If you just use density directly on width and height, you get the values that look the same as on the device, at least on Pixel4a. I cannot confirm on other phones though...

takahirom: Thank you. I didn't notice that. I fixed it by using pixelDensity. For instance, the value for AndroidTV1080p was w7670dp, and now it's w960dp. It seems correct.
https://github.com/takahirom/roborazzi/pull/49/commits/a797fcccd7cf3d5c4ee81cc1ab8c4eb1393e5da0

sergio-sastre (approved):

roborazzi/src/main/java/com/github/takahirom/roborazzi/RobolectricDeviceQualifiers.kt
```diff
@@ -0,0 +1,30 @@
...
+object RobolectricDeviceQualifiers {
+  // Generated by [GenerateQualifiersTest.kt]
+  // Data from: AOSP https://android.googlesource.com/platform/tools/base/+/mirror-goog-studio-master-dev/sdklib/src/main/java/com/android/sdklib/devices
+// FROM: tv.xml
+  const val AndroidTV1080p = "w960dp-h540dp-xlarge-long-notround-television-xhdpi-keyshidden-dpad"
+  const val AndroidTV720p = "w962dp-h541dp-xlarge-long-notround-television-tvdpi-keyshidden-dpad"
+
+  // FROM: devices.xml
+// FROM: wear.xml
+  const val WearOSSquare = "w187dp-h187dp-small-long-round-watch-hdpi-keyshidden-nonav"
```
yschimke, line 12: I'll put up a PR to replace WearOSRound and WearOSRoundChin with WearOSLargeRound and WearOSSmallRound.

yschimke: Done in https://github.com/takahirom/roborazzi/pull/56

Issue #50, opened 2023-04-25
Feature request: provide PNG of actual result when verifying


alexjlockwood:
For example, when verifying it seems like there are two things that end up in the `build/outputs/roborazzi` directory:

- The original image that was recorded in a previous "record" step 
- A comparison image that compares the recorded image above with the new/actual image

It would be nice if the "new/actual" image could also be provided separately. This way we can easily use it as a new baseline recorded image if we wish (i.e. if the changes in the test are actually intentional and we want to update our "golden value" image to use the next time the test is run).

takahirom:
Thanks. I would like to clarify your request. The new image will be saved as "{file_name}_actual.png". Is this what you are looking for?

takahirom:
If you don't mind, can you take a look at this PR? https://github.com/takahirom/roborazzi/pull/51

Issue #23, opened 2023-04-16
Seeking guidance in integrating an isolated example   


TonyTangAndroid:
Hi there,

I have been playing around this library by checking out the whole repo locally and running it locally. I also made effort in creating a full isolated app to play around the library.

However, I could not figure out the effectiveness of this library.
To start with, I could not find any snapshot view generated under `build/outputs/roborazzi` .
Then , how would it catch a bug if I purposefully break the code?   Running `./gradlew recordRoborazziDebug` or ` ./gradlew verifyRoborazziDebug`  does not break at all matter how I change the code.

I think there must be gap that I failed to interpret it by following the readme.
 

I tried to find any other sample app that is using `Roborazzi`. But I only find `coil` library, which is large. Hence, I did not explore it further. 
Could you kindly check out this sample here  https://github.com/TonyTangAndroid/HelloRoborazzi and see to the gap?

It would be great that we could have a minimum app to demonstrate how `Roborazzi` is working.

takahirom:
Thank you for your feedback and for providing the sample code. I checked, and it seems that your robo module was not applied with the roborazzi plugin. To take a screenshot, we need to use the plugin.
I have created this PR:
https://github.com/TonyTangAndroid/HelloRoborazzi/pull/3

Please run:
./gradlew robo:recordRoborazziDebug

As for the test not failing, we are currently only generating images for comparison, not causing the test to fail. I plan to address this issue by separating the compare and verify Gradle tasks in another PR.

takahirom:
As far as samples go, I believe the app module in this repository is currently reasonably simple. Please let me know if there are other problems and I will open this issue again.

TonyTangAndroid:
I see.  I think I know the gap now. 
`we are currently only generating images for comparison, not causing the test to fail.`.

Thank you for the clarification. 

I had the wrong assumption that since it is built for test purpose, hence it will fail if I introduce regression. And I had tried it out all the ways but failed to make it fail.
I forked this library and purposefully added a breaking change. Then it turned out that it still passed the CI. https://github.com/TonyTangAndroid/roborazzi/pull/2/files 

Looking forward to the feature  `separating the compare and verify Gradle tasks in another PR.`

TonyTangAndroid:
I figured it out the baseline now. Will revise my sample app to make it simpler!

takahirom:
Thank you for trying it out! Please let me know if there is anything else for improvement.

alexjlockwood:
[...]

PR #51, opened 2023-04-26
Save the actual image when comparing and verifying

takahirom:
https://github.com/takahirom/roborazzi/issues/50


roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -547,6 +547,12 @@ private fun saveOrCompare(
   goaldenFile: File,
   roborazziOptions: RoborazziOptions
 ) {
+  val forbiddenFileSuffixes = listOf("_compare", "_actual")
+  forbiddenFileSuffixes.forEach {
+    if (goaldenFile.nameWithoutExtension.endsWith(it)) {
```
yschimke, line 552: nit: 
```suggestion
    if (goldenFile.nameWithoutExtension.endsWith(it)) {
```

Bencodes:
@takahirom are you still planning on merging this one?

takahirom:
@Bencodes 
I haven't merged the change yet because I'm not confident that it's exactly what you're looking for. Could you please confirm if the proposed solution works for you?

Bencodes:
> Could you please confirm if the proposed solution works for you?

The behavior that we are looking for is to be able to record, verify against the existing baseline, and export the new baseline (even if they changed) in a single pass so that we can copy out the updated baseline images when the user decides to make them the new baselines. Having the actuals available alongside the original baselines also makes it easier to post the images directly to Github as a comment for comparison without having to do subsequent builds to get the new snapshots out.

takahirom:
@Bencodes 
Thanks for your feedback. I can add a verifyAndRecordRoborazziDebug task, but there is an issue when a single test contains multiple captureRoboImage() calls and each of them is different. In this case, the first mismatch would cause an assertion, preventing the rest of the images from being updated. This behavior is designed to easily identify and resolve problems using the stack trace of the location where the issue occurred. Is this still acceptable to you?
https://github.com/takahirom/roborazzi/issues/23#issuecomment-1515090642

takahirom:
Even if we need to rerun the task, implementing verifyAndRecordRoborazziDebug might be a good choice. This is because it adheres to the best practice of keeping tests separate. Also, even in the context of a scenario test, this approach can be beneficial because if multiple images are saved, it can become difficult to pinpoint the cause of failure.

Bencodes:
> This behavior is designed to easily identify and resolve problems using the stack trace of the location where the issue occurred. Is this still acceptable to you?

I think this makes sense. And I agree that in terms of best practices one-snapshot-per-test is the correct approach.

takahirom:
Regardless, I'm planning to add the implementation to this PR. So, I'll go ahead and merge this for now.

Issue #39, opened 2023-04-20
Error `Iteration already started` when performing click and using Roborazzi Rule


bohregard:
I have a fairly simple test that presses some buttons to toggle between weeks and update some text to reflect the new week.
When I add Roborazzi, the test fails with the error message `Iteration already started`. I've also noticed that the test times are a lot longer than without Roborazzi (see attached images).

```
@RunWith(RobolectricTestRunner::class)
@GraphicsMode(GraphicsMode.Mode.NATIVE)
class CalendarUiTests {
    @get:Rule
    val composeTestRule = createAndroidComposeRule<ComponentActivity>()

    @get:Rule
    val roborazziRule = RoborazziRule(
        composeRule = composeTestRule,
        captureRoot = composeTestRule.onRoot(),
        options = RoborazziRule.Options(
            captureType = RoborazziRule.CaptureType.LastImage,
        )
    )

    @Before
    @Throws(Exception::class)
    fun setup() {
        ShadowLog.stream = System.out
    }

    @Test
    fun `calendar top bar changes dates`() {
        var start by mutableStateOf(LocalDateTime.of(
            LocalDate.of(2023, 10, 1),
            LocalTime.MIDNIGHT
        ))
        var end by mutableStateOf(LocalDateTime.of(
            LocalDate.of(2023, 10, 7),
            LocalTime.MAX
        ))

        composeTestRule.setContent {
            CalendarTopBar(
                start = start,
                end = end,
                onPreviousWeek = {
                    start = start.minus(7, ChronoUnit.DAYS)
                    end = end.minus(7, ChronoUnit.DAYS)
                },
                onNextWeek = {
                    start = start.plus(7, ChronoUnit.DAYS)
                    end = end.plus(7, ChronoUnit.DAYS)
                }
            )
        }

        composeTestRule.onRoot().printToLog("TAG")

        composeTestRule.onNodeWithTag("calendarTitleBar")
            .assertIsDisplayed()
            .assertTextContains("Oct 1 - Oct 7")

        composeTestRule.onNodeWithTag("buttonPreviousWeek")
            .performClick()

        composeTestRule.onRoot().printToLog("TAG")

        composeTestRule.onNodeWithTag("calendarTitleBar")
            .assertIsDisplayed()
            .assertTextContains("Sep 24 - Sep 30")

        composeTestRule.onNodeWithTag("buttonNextWeek")
            .performClick()
            .performClick() // Fails on this line

[...]

PR #16, opened 2023-03-27
[WIP]Use DrawListener for NativeGraphics

takahirom:
OnDrawListener seems it does not work now

PR #52, opened 2023-04-28
Fix Compose gif

takahirom:
It seems that ViewTreeObserver.OnGlobalLayoutListener doesn't work when using Compose. So, I tried to use [Snapshot.registerApplyObserver{}](https://dev.to/zachklipp/introduction-to-the-compose-snapshot-system-19cn).

https://github.com/takahirom/roborazzi/pull/16#issuecomment-1525874734

Fix https://github.com/takahirom/roborazzi/issues/39

### DIFF
--- a/app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeSetContentTest.kt
+++ b/app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeSetContentTest.kt
@@ -1,0 +1,90 @@
+package com.github.takahirom.roborazzi.sample
+
+import androidx.activity.ComponentActivity
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.unit.dp
+import com.github.takahirom.roborazzi.RobolectricDeviceQualifiers
+import com.github.takahirom.roborazzi.RoborazziRule
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/github/takahirom/roborazzi/sample/ComposeSetContentTest.kt
```diff
@@ -0,0 +1,90 @@
...
+  val roborazziRule = RoborazziRule(
+    composeRule = composeTestRule,
+    captureRoot = composeTestRule.onRoot(),
+    options = RoborazziRule.Options(
+      RoborazziRule.CaptureType.Gif
+    )
+  )
+
+  @Test
+  fun composable() {
```
takahirom, line 51: com_github_takahirom_roborazzi_sample_ComposeSetContentTest_composable

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -460,30 +479,26 @@ fun SemanticsNodeInteraction.captureComposeNode(
...
+      canvases.addIfChanged(it, roborazziOptions)
+    }
+  }
+  val handler = Handler(Looper.getMainLooper())
+  val composeApplyObserver = Snapshot.registerApplyObserver { anies, snapshot ->
+    handler.postAtFrontOfQueue {
+      try {
+        capture()
+      } catch (e: Exception) {
+        // It seems there is no screen, so wait
```
yschimke, line 491: This seems quite unspecific, could it fail for other reasons?

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -429,6 +432,24 @@ fun ViewInteraction.captureAndroidView(
...
+  roborazziOptions: RoborazziOptions
+) {
+  val prev = this.lastOrNull() ?: run {
+    this.add(next)
+    return
+  }
+  val differ: ImageComparator.ComparisonResult =
+    prev.differ(next, 1.0)
+  if (!roborazziOptions.compareOptions.resultValidator(differ)) {
+    this.add(next)
```
yschimke, line 446: Should this be configurable? I'd like to be able to use this for capturing frame by frame animations.

yschimke (approved):
Looks great, couple of questions.

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
takahirom: You have a point, but I think this implementation is better than the current one. I'll create an issue for this. 👍

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
takahirom: Good point. I fixed the comments, and it no longer uses Exception for handling exceptions.

https://github.com/takahirom/roborazzi/pull/52/commits/6045cfcd2a4fdf8a4a4f7d8d48ad62d0d0255869

Issue #46, opened 2023-04-22
Add A11y transform


yschimke:
For inspiration https://github.com/joreilly/Confetti/blob/de9c9cd1f309071a44f6879c637e895770573a81/androidTest/src/main/kotlin/dev/johnoreilly/confetti/screenshot/a11y/A11ySnapshotTransformer.kt#L17

Results in screenshots like https://github.com/joreilly/Confetti/blob/de9c9cd1f309071a44f6879c637e895770573a81/wearApp/snapshot/ConferenceHomeScreenTest/conferenceHomeScreenA11y_Colors_Material__reference.png

![image](https://user-images.githubusercontent.com/231923/233776319-a54c0978-736a-44d7-a36d-eb326c8b1460.png)

takahirom:
We have a dump feature in this library. How about customizing it for accessibility (A11y)?
https://github.com/takahirom/roborazzi#dump-mode

takahirom:
@yschimke 
I've implemented it. Could you take a look?
https://github.com/takahirom/roborazzi/pull/55

PR #55, opened 2023-05-01
Add accessibility dump

takahirom:
## Why

https://github.com/takahirom/roborazzi/issues/46

## How

```kotlin
val explanation: ((RoboComponent) -> String)
```

```kotlin
    onView(ViewMatchers.isRoot())
      .captureRoboImage(
        filePath = "$DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH/manual_view_first_screen_with_query_compose_custom.png",
        roborazziOptions = RoborazziOptions(
          captureType = RoborazziOptions.CaptureType.Dump(
            explanation = RoborazziOptions.CaptureType.Dump.AccessibilityExplanation, // ←
          )
        )
      )
```

image


takahirom:
👀 
https://github.com/google/talkback/blob/master/utils/src/main/java/com/google/android/accessibility/utils/TreeDebug.java

takahirom:
It seems that AccessibilityNodeInfoCompat.obtain(view) doesn't work well. 👀
If I use nodeDebugDescription from [TreeDebug.java](https://github.com/google/talkback/blob/master/utils/src/main/java/com/google/android/accessibility/utils/TreeDebug.java), it will be like this:
image

takahirom:
However, I believe it is better to have some implementation, even if it is not perfect.

yschimke:
This looks great. Building the accessibility description in makes sense.  The labels with lines look great.

I'd love to be able to provide a different implementation, not using the exploded view, but just showing a flat image with highlighted rectangles.  Is that possible to make overridable?

Maybe it's just changing the elevation (to zero), and the layer transparency (to fully)?

yschimke:
What's with the "-1??" text in the second image?

takahirom:
There is a depthSlideSize property that allows you to change the slide size. As for layer transparency, we don't currently have that option. However, I believe we can add it.

https://github.com/takahirom/roborazzi/blob/main/roborazzi/src/main/java/com/github/takahirom/roborazzi/capture.kt#L210C11-L210C25

It might be the window ID. I agree that it's not very readable. Regardless, I'm not going to use the TalkBack implementation in this PR.

https://github.com/takahirom/roborazzi/commit/4301660a4bc5836e9665e98db4c5e05a406e5e42#diff-b5721281cee7bf3e6758c3f8943ed7bed6de8caa71dfc072a2d22752c3549126R63

takahirom:
Add some properties
image

PR #64, opened 2023-05-02
Add an IDE icon.

yschimke:
https://commons.wikimedia.org/wiki/File:Twemoji12_1f916.svg

This file is licensed under the Creative Commons Attribution 4.0 International license.


takahirom:
Thank you. I tried to generate a logo using AI, but I couldn't find any image that I wanted to use. So, for now, it's good to use the icon!

PR #69, opened 2023-05-04
Refactor and reduce Gradle Plugin logic

takahirom:


takahirom:
The screenshot test found the actual bug 😄

takahirom:
The failure of `unchanged check test / test (pull_request)` is OK. It is expected to fail.

Issue #67, opened 2023-05-03
White background on component images


yschimke:
Existing screenshots before roborazzi 

<img width="488" alt="image" src="https://user-images.githubusercontent.com/231923/235940566-278f3d40-3fb7-4348-a305-0c6404b80e31.png">

After roborazzi https://github.com/google/horologist/pull/1234

<img width="486" alt="image" src="https://user-images.githubusercontent.com/231923/235940864-809792a3-ec30-4d9f-8e58-0d155725d4db.png">

takahirom:
How did you take this capture? Did you perhaps use this method? Maybe we should use a transparent Activity theme 👀 
https://github.com/takahirom/roborazzi/blob/main/roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt#L187

yschimke:
Test is here https://github.com/google/horologist/blob/61d33dfd081c37d83fe063a46f73d06b9e6ea88e/base-ui/src/test/java/com/google/android/horologist/base/ui/components/PrimaryChipTest.kt

```
@RunWith(AndroidJUnit4::class)
@GraphicsMode(GraphicsMode.Mode.NATIVE)
@Config(
    sdk = [30],
    qualifiers = RobolectricDeviceQualifiers.WearOSLargeRound
)
class PrimaryChipTest {
    @get:Rule
    val composeTestRule = createAndroidComposeRule<ComponentActivity>()

    @get:Rule
    val roborazziRule = RoborazziRule(
        composeRule = composeTestRule,
        captureRoot = composeTestRule.onRoot(),
        options = RoborazziRule.Options(
            captureType = RoborazziRule.CaptureType.LastImage,
            outputDirectoryPath = "src/test/snapshots/images",
            roborazziOptions = RoborazziOptions(
                recordOptions = RoborazziOptions.RecordOptions(
                    pixelBitConfig = RoborazziOptions.PixelBitConfig.Rgb565
                )
            )
        )
    )
```

I've just realised I'm not setting a Theme, on Wear, it should default to black.  Which would be better, but I'd prefer transparent.

Just realised it might be because I'm not using ARGB_8888

yschimke:
Still happening with ARGB_8888

takahirom:
@yschimke 
I fixed the transparency issue in the Compose lambda. Could you take a look at it?
[...]

PR #70, opened 2023-05-04
Use transparent theme for compose lambda

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/67

### DIFF
--- a/roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
+++ b/roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
@@ -12,7 +12,6 @@ import android.os.Looper
 import android.view.View
 import android.view.ViewGroup
 import android.view.ViewTreeObserver
-import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.snapshots.Snapshot
@@ -187,7 +186,7 @@ fun captureRoboImage(
   content: @Composable () -> Unit,
 ) {
   if (!roborazziEnabled()) return
-  val activityScenario = ActivityScenario.launch(ComponentActivity::class.java)
+  val activityScenario = ActivityScenario.launch(RoborazziTransparentActivity::class.java)
   activityScenario.onActivity { activity ->
     activity.setContent {
       content()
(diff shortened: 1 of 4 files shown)


takahirom:
The failure of `unchanged check test / test (pull_request)` is OK. It is expected to fail.

yschimke (approved):
LGTM - although I wonder if it breaks some apps currently assuming a White background.  Coming from Wear, I think that's ok, as it wouldn't previously have a black background.

yschimke:
It might be useful to demonstrate how to get back to a Theme background, should this be an option?

takahirom:
I think the new Compose function hasn't been widely used yet. People can create their own custom captureRoboImage(). I'm going to mention this PR comment in the release notes.

```kotlin
fun captureRoboImageNoTransparent(
  file: File,
  roborazziOptions: RoborazziOptions = RoborazziOptions(),
  content: @Composable () -> Unit,
) {
  if (!roborazziEnabled()) return
  // Use non-transparent activity
  val activityScenario = ActivityScenario.launch(ComponentActivity::class.java)
  activityScenario.onActivity { activity ->
    activity.setContent {
      content()
    }
    val composeView = activity.window.decorView
      .findViewById<ViewGroup>(android.R.id.content)
      .getChildAt(0) as ComposeView
    val viewRootForTest = composeView.getChildAt(0) as ViewRootForTest
    viewRootForTest.view.captureRoboImage(file, roborazziOptions)
  }
}
```

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -187,7 +186,7 @@ fun captureRoboImage(
   content: @Composable () -> Unit,
 ) {
   if (!roborazziEnabled()) return
-  val activityScenario = ActivityScenario.launch(ComponentActivity::class.java)
+  val activityScenario = ActivityScenario.launch(RoborazziTransparentActivity::class.java)
```
takahirom, line 189: image

takahirom:
However, it seems difficult to view images without a background color. We are not currently using background colors in other parts, and users can use Rgb565 as a workaround. Therefore, I believe it is a good idea to use transparency in this PR.
image

yschimke:
Transparent works for me, once this is allowed you can choose any other strategy.  But critically if transparent isn't supported, it's hard to put it back later.

I actually found the background was the big difference between Screen and Component level screenshots. Maybe something you could support different strategies, Fill, 50PctAlpha, Transparent, DeviceTheme?

For example

https://github.com/google/horologist/blob/main/roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt#L248

```
    @Composable
    private fun ComponentDefaults(
        fakeImageLoader: FakeImageLoader,
        content: @Composable () -> Unit
    ) {
        fakeImageLoader.override {
            Box(
                modifier = Modifier
                    .wrapContentSize()
                    .background(Color.Black.copy(alpha = 0.5f)),
                contentAlignment = Alignment.Center
            ) {
                DefaultTheme {
                    content()
                }
            }
        }
    }
```

yschimke (approved):

PR #81, opened 2023-05-12
Add verifyAndRecordRoborazzi task

takahirom:
This PR introduces a new task, verifyAndRecordRoborazzi, which aims to streamline the process of recording and verifying tests into a single run. This approach enhances efficiency, although it may require multiple runs if a single test contains multiple captureRoboImage() calls when the images are changed. Feedback is welcome.

See https://github.com/takahirom/roborazzi/pull/51#issuecomment-1543442519 for more information.

### DIFF
--- a/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
+++ b/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
@@ -23,6 +23,7 @@ class RoborazziPlugin : Plugin<Project> {
     val verifyVariants = project.tasks.register("verifyRoborazzi")
     val compareVariants = project.tasks.register("compareRoborazzi")
     val recordVariants = project.tasks.register("recordRoborazzi")
+    val verifyAndRecordVariants = project.tasks.register("verifyAndRecordRoborazzi")
 
     val hasLibraryPlugin = project.pluginManager.hasPlugin("com.android.library")
     val variants = if (hasLibraryPlugin) {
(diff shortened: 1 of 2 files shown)


takahirom:
@yschimke @Bencodes @alexjlockwood
Could you please take a look at this PR?

takahirom:
While it is possible to execute ./gradlew verifyRoborazziDebug recordRoborazziDebug, creating a reliable task for this use case can be problematic because communicating between tasks can be difficult, which is why we opted for a different approach.

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -591,7 +591,7 @@ private fun saveOrCompare(
       true
```
Bencodes, line 591: Does it make sense to rename the function `saveOrCompare` wrapping this logic given it could be saving and comparing at the same time?

takahirom: Thank you for your suggestion. I agree, the name 'processOutputImageAndReport' more accurately represents the function's behavior as it can both save and compare images. I will make this change.

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -609,10 +609,15 @@ private fun saveOrCompare(
...
-      val actualFile = File(
-        goaldenFile.parent,
-        goaldenFile.nameWithoutExtension + "_actual." + goaldenFile.extension
-      )
+      val actualFile = if (roborazziRecordingEnabled()) {
+        // If record option is enabled, we should save the actual file as the golden file.
+        goaldenFile
+      } else {
+        File(
+          goaldenFile.parent,
```
Bencodes, line 617: What ends up happening if you are using `VerifyAndRecord` without a baseline to feed in?

roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -23,6 +23,7 @@ class RoborazziPlugin : Plugin<Project> {
     val verifyVariants = project.tasks.register("verifyRoborazzi")
     val compareVariants = project.tasks.register("compareRoborazzi")
     val recordVariants = project.tasks.register("recordRoborazzi")
+    val verifyAndRecordVariants = project.tasks.register("verifyAndRecordRoborazzi")
```
Bencodes, line 26: Looks fine but we don't use Gradle so I'd have someone else review if you think it needs a set of eyes.

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
takahirom: In that scenario, the test fails and generates an output image file and comparison file. I believe this is an inevitable result when we use the verify task, considering its purpose.

roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
takahirom: Thanks. I don't think it's a big change to the Gradle tasks. So we can fix it if there's anything.

takahirom:
Thanks. I'm about to release these changes. If you notice anything, please let me know so I can fix it. 🙏

Issue #80, opened 2023-05-10
Option to change file naming strategy


takahirom:
I initially thought it would be good to use "_" instead of "." in the file names. However, I'm now beginning to think it might be better to leave the names as they are because the JUnit report's file name will remain unchanged. This could be a breaking change, and we may need to provide some compatibility options.

If you are okay with a quick fix even if it's a breaking change, please react with 🚀 
If you would like it to be an option but don't mind a slower fix, please react with 🎉 
If you think no changes are needed, please react with 👀

Issue #75, opened 2023-05-07
Allow setting default values for Roborazzi options and gather suggestions


takahirom:
We may want to pass parameters like resizeScale or pixelBitConfig.
I'm going to add some options as system properties. You can pass parameters like this: "./gradlew recordRoborazziDebug -PresizeScale=0.5"

Are there any parameters you'd like to set as default?

PR #83, opened 2023-05-15
Add default options parameters like "roborazzi.record.resizeScale" and "roborazzi.record.namingStrategy"

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/80 , https://github.com/takahirom/roborazzi/issues/75

This will constitute a breaking change. The default filename will be altered to something like `com.github.takahirom.roborazzi.sample.ManualTest.captureRoboImageSample.png` from `com_github_takahirom_roborazzi_sample_ManualTest.captureRoboImageSample.png`

You can try it like this.

./gradlew recordRoborazziDebug -Proborazzi.record.resizeScale=0.5 -Proborazzi.record.nameStrategy=testClassAndMethod

or

gradle.properties

```
roborazzi.record.resizeScale=0.25
roborazzi.record.namingStrategy=testClassAndMethod
```

https://github.com/takahirom/roborazzi/pull/84

For compatibility, you can use `escapedTestPackageAndClassAndMethod `. But keep in mind it can be deprecated.

```
roborazzi.record.namingStrategy= escapedTestPackageAndClassAndMethod
```

### DIFF
--- a/.github/workflows/compare-verify-test.yaml
+++ b/.github/workflows/compare-verify-test.yaml
@@ -40,7 +40,7 @@ jobs:
           ./gradlew app:compareRoborazziDebug --stacktrace
           # Check if there are difference with app/build/outputs/roborazzi/test_compare.png
           find app/build/outputs/roborazzi
-          file_path="app/build/outputs/roborazzi/com_github_takahirom_roborazzi_sample_ManualTest_captureRoboImageSample_compare.png"
+          file_path="app/build/outputs/roborazzi/com.github.takahirom.roborazzi.sample.ManualTest.captureRoboImageSample_compare.png"
           if [ ! -f "$file_path" ]; then
             echo "Error: File $file_path not found."
            exit 1
--- a/roborazzi/src/main/java/com/github/takahirom/roborazzi/DefaultFileNameGenerator.kt
+++ b/roborazzi/src/main/java/com/github/takahirom/roborazzi/DefaultFileNameGenerator.kt
@@ -8,8 +8,7 @@ object DefaultFileNameGenerator {
   enum class DefaultNamingStrategy(val optionName: String) {
     TestPackageAndClassAndMethod("testPackageAndClassAndMethod"),
     EscapedTestPackageAndClassAndMethod("escapedTestPackageAndClassAndMethod"),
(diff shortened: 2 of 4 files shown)


takahirom:
@yschimke, I've added a 'roborazzi.record.namingStrategy' option to Gradle. I'd be interested in hearing your thoughts on this.

yschimke:
I'm not sure about using it as a gradle command line option. Things will break if you don't set it, or set incorrectly. So config makes more sense to me. 

The main reason I want this to be exposed is for migrating from an existing project structure. So I'd still hope it's extensible, so I can match paths.

takahirom:
@yschimke 
Thank you for your input. I understand your concerns. However, I believe we can add these options to gradle.properties.
https://github.com/takahirom/roborazzi/pull/84/files#diff-3d103fc7c312a3e136f88e81cef592424b8af2464c468116545c4d22d6edcf19R57-R59

As for 'matching paths', could you please elaborate? Are you referring to something like 'com/example/Main.png'?

yschimke:
Yep, I'm sure it will work, I have a bias towards type safe config in build.gradle.kts over gradle.properties.

On paths DefaultNamingStrategy seems restrictive, I don't use Paparazzi and my current RNG based solution in the same way as you are suggesting.  

I think it's ok because I can override here

```
fun ViewInteraction.captureRoboImage(
  filePath: String = DefaultFileNameGenerator.generateFilePath("png"),
```

But ideally I could just set a PaparazziFilenameConventions and have it match by default.

But I can work with this.

takahirom:
@yschimke 
I understand your preference for type-safe configuration in build.gradle.kts over gradle.properties.

Certainly, we can add an extension later. This current implementation could serve as the foundation for it.

Regarding the path issue, we could introduce a DefaultNamingStrategy. While this implementation might not completely satisfy your needs, it could serve as a stepping stone towards a more suitable solution.

yschimke:
Yep, no issue with the PR. You have to balance a simple API with random requests by users. Just giving my opinion.

I can work with this.

takahirom:
Indeed, it can make the change challenging. I plan to customize the API to suit my needs.

PR #87, opened 2023-05-19
Modernize AGP integration

ZacSweers:
This updates the gradle plugin to use modern AGP components APIs for reacting to variants. This allows not relying on `internal` packages and also fixes the plugin's race condition that requires it to be applied after the android plugin.


takahirom:
👀

ZacSweers:
Will look at the CI failures tomorrow!

ZacSweers:
Sorry I don't know how to test this locally because it's impossible to build the plugin if the downstream apps can't configure with it :/

takahirom:
Thank you for trying to fix it.
Here are the commands I usually use to check the Gradle plugin. In this way, I can reload the plugin implementation that the app refers to.

```
cd roborazzi
rm roborazzi-gradle-plugin/build/libs/roborazzi-gradle-plugin.jar ; ./gradlew app:test --dry-run ; ./gradlew roborazzi-gradle-plugin:build ; ./gradlew app:recordRoborazziDebug
```

ZacSweers:
I see. Would you be up for me changing this to use a more conventional included build setup so it doesn't require building jars manually first? Similar to what we do here: https://github.com/slackhq/keeper

takahirom:
I attempted to apply the "include build" function to Roborazzi. However, a significant challenge that needs to be addressed is that the roborazzi-core module is utilized by both the plugin and libraries. I'm at a loss for a solution at the moment.
https://github.com/takahirom/roborazzi/pull/88

takahirom:
I didn't understand include build well. I'm going to try dependencySubstitution for the module.  👀

takahirom:
@ZacSweers 
I've updated the plugin to use the 'includeBuild' feature in the main branch. Could you check if it works for you?

https://github.com/takahirom/roborazzi/blob/main/settings.gradle#L22-L27

takahirom:
I've pushed to this branch to resolve the conflict. If you disagree with my changes, please feel free to force push.

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -24,92 +25,98 @@ class RoborazziPlugin : Plugin<Project> {
...
+        project.gradle.taskGraph.whenReady { graph ->
+          isRecordRun.set(recordTaskProvider.map { graph.hasTask(it) })
+          isVerifyRun.set(verifyTaskProvider.map { graph.hasTask(it) })
+          isVerifyAndRecordRun.set(verifyAndRecordTaskProvider.map { graph.hasTask(it) })
+          isCompareRun.set(compareReportGenerateTaskProvider.map { graph.hasTask(it) })
         }
+
+        val testTaskProvider = project.tasks.withType(Test::class.java)
+          .matching { it.name == "test$testVariantSlug" }
+          .configureEach { test ->
```
takahirom, line 81: Perhaps this function returns Unit, and we are using the testTaskProvider variable as a parameter for dependsOn(). This appears to be the cause of the error we are seeing. 👀

takahirom:
Now that I have lots of tests for this. I think I can merge this 🙏

ZacSweers:
Nice!

PR #105, opened 2023-06-20
Add support for default file path setting by RoborazziRule and refactor RoborazziRule

### LINKED ISSUE #104: Single compose test always fails locally, does not fail CI verify call even when snapshots are missing
Repro: https://github.com/slackhq/circuit/pull/669

Run the `PetListSnapshotTest.petList_show_list_for_success_state()` test, observe that it always fails locally and cannot record a new snapshot. On CI however, the verify calls always pass despite there being no available snapshot.

This is using 1.3.0-alpha-4

takahirom:
Fix #104 

This PR introduces an API to set the default file path within the Roborazzi context. This enables the client to override the default file path as needed and clear it after use.

This PR also fixes the CaptureType.LastImage behavior. It should not utilize the mechanism for generating GIFs, as this leads to the creation of an excessive number of images.


takahirom:
Oh, It breaks wear image 👀

Issue #97, opened 2023-06-02
--rerun-tasks Required for Saving Baseline Using Cache


galisamas:
Hey, 

I am trying to run Roborazzi snapshots in the pipeline which is configured on Jenkins. However, I am facing a few issues (or maybe this is something I misunderstood in the setup process and you will be able to guide me). There are two issues that I want to stress, which might be related:

Prerequisites: 
- I am running the `1.2.0` version
- Have baseline stored in the project
- I am using the most simple Roborazzi Rule setup. `RoborazziRule(onView(isRoot()))`

1. Running verifyRoborazziDebug task in Jenkins.
No matter what `path` I will declare in the `captureRoboImage(path)` for the snapshot using, it will always fail. However, this doesn't apply to the local gradle task run. I have a test that should be using a custom `path` to record and verify the snapshot state. It works locally, but in Jenkins, I always get this:
```
<...>MyTest > example1 FAILED
    java.lang.AssertionError: Roborazzi: /var/lib/jenkins/workspace/<...>/build/outputs/roborazzi/example1_compare.png is added.
    See compare image at /var/lib/jenkins/workspace/<...>/build/outputs/roborazzi/example1_compare.png
        at com.github.takahirom.roborazzi.RoborazziOptions$RoborazziCompareReporter$VerifyRoborazziCompareReporter.report(capture.kt:349)
        at com.github.takahirom.roborazzi.RoborazziKt.processOutputImageAndReport(Roborazzi.kt:666)
        at com.github.takahirom.roborazzi.RoborazziKt.saveLastImage(Roborazzi.kt:472)
        at com.github.takahirom.roborazzi.RoborazziKt.access$saveLastImage(Roborazzi.kt:1)
        at com.github.takahirom.roborazzi.RoborazziKt$captureAndroidView$3.invoke(Roborazzi.kt:432)
        at com.github.takahirom.roborazzi.RoborazziKt$captureAndroidView$3.invoke(Roborazzi.kt:431)
        at com.github.takahirom.roborazzi.RoborazziRule$apply$1.evaluate(RoborazziRule.kt:133)
        at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
        at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
        at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
[...]

PR #116, opened 2023-07-09
Fix issue test task is skippped unexpectedly

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/97


takahirom:
Anyway, I believe this is better than the current implementation.

Issue #125, opened 2023-07-19
Allow separate directories for golden screenshots & comparison images


simonwalter:
Is there a way to specify separate directories for the golden screenshots and the comparison images?

### Background
In our project, we use the `outputDirectoryPath` option to specify where golden screenshots should be stored. Currently, we specify a directory outside of `/build` and commit the generated files to Git using Git LFS. Here's an example:

```kotlin
class RuleTestWithPath {
  @get:Rule
  val roborazziRule = RoborazziRule(
    options = Options(
      outputDirectoryPath = "screenshots/custom_outputDirectoryPath",
      ...
    ),
  )
```
![image](https://github.com/takahirom/roborazzi/assets/5316135/f0b02dc9-bebc-4818-9b8e-02be10ca68be)

However, one limitation of using `outputDirectoryPath` is that the images generated by `verifyRoborazziDebug` for failing tests will also be put into the same folder: 

![image](https://github.com/takahirom/roborazzi/assets/5316135/caad893f-6ef2-4a95-a01d-22450497ca74)

While keeping the golden screenshots stored in the path specified by `outputDirectoryPath`, we would prefer to have the comparison images located inside the `/build` folder. This approach offers the following advantages:
- It helps to prevent accidental commits of the comparison files to Git, eliminating the need for manual deletion of the comparison files.
- When golden screenshots and comparison files are already separated, it becomes easier to store only the comparison images in our CI system for reviewing each run.

takahirom:
Thank you for your suggestion. As a temporary workaround, you might be able to use the "_compare.png" naming convention that Roborazzi automatically applies to the generated files. These files could be moved to the desired location using a hook in the build task or something similar.

For more effective management of the golden screenshots and comparison files, what are your thoughts on introducing an additional option such as comparisonOutputPath to the Rule, alongside outputDirectoryPath? This wouldn't automate the relocation of files, but it could assist in specifying distinct paths for the golden screenshots and the comparison images.

takahirom:
I'm considering extending the FileProvider, but it could lead to a rather complex implementation. As an alternative, what if we create a FileOutputInterceptor as a parameter to the Rule, like this?

```kotlin
sealed interface FileType {
  class Reference : FileType
  class Comparison : FileType
  class Actual : FileType
}

fun interface OutputFileInterceptor {
  fun intercept(fileType: FileType, original: File): File {

  }
}
```

simonwalter:
[...]

PR #128, opened 2023-07-22
Alter the default output directory for comparison images and incorporate outputDirectoryPath into the comparison options

takahirom:
#125


blundell:
👍  I was just reading the discussion in #125 and I was expecting this PR to be able to set 2 different output directories? One for the compare task and a different dir for the record task. Is this not the case? (only see changes in `CompareOptions`)

blundell:
Think I've worked it out, the `filePath` param on the `captureRoboImage` method is the record file path.
I assumed it'd be in `RecordOptions` 😅

PR #129, opened 2023-07-24
Add message in exception if image was not found

blackbracken:
# About
When taking a screenshot, if `rootComponent.image` is `null`, throw an exception with the message "Does the rendering element exist?" instead of an NPE.

# Details
In the implementation of `RoboComponent.image` (View, Compose), both ultimately call `View.fetchImage` to attempt to obtain a bitmap, and the image is null only if the size is less than 0 or the bitmap generation fails. If the cause of the bitmap generation failure (i.e. `bitmapFuture` returning `null`) during screenshot is extremely rare, the user most likely has not placed any rendering elements.

This is not something that can be easily read from the stack trace (for beginners), so I will add a message to inform about it.

This is something I felt would be good to have in roborazzi when I encountered a case where I didn't render anything in Compose.

### DIFF
--- a/roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
+++ b/roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
@@ -703,7 +703,8 @@ internal fun capture(
     )
 
     is RoborazziOptions.CaptureType.Screenshot -> {
-      val image = rootComponent.image!!
+      val image = rootComponent.image
+        ?: throw IllegalStateException("Unable to find the image of the target root component. Does the rendering element exist?")
       onCanvas(
         RoboCanvas(
           width = image.width,


takahirom:
👀

takahirom (approved):
Looks great!

ashughes:
I'm running into this exception. I've followed the steps to set up roborazzi and generate compose screenshots. When I run `recordRoborazziDebug`, it takes a very long time (over 24m on an M2 Macbook Pro) and all the `RoborazziPreviewParameterizedTests` fail with the `IllegalStateException` added in this PR. I'm not sure if this is a bug or if there's something I'm missing with the setup that would cause this exception. I'm not really sure where to go from here, so any pointers would be appreciated! Happy to file a separate bug if necessary.

takahirom:
Thank you for your question.

Here’s what I would do: Use the debugger to run the test and pause execution at the point where it gets stuck to investigate the cause.

1. First, set up your test to ensure the screenshot test runs by adding the following configuration:
roborazzi.test.record=true
https://github.com/takahirom/roborazzi?tab=readme-ov-file#q-how-can-i-debug-screenshot-tests-in-android-studio


2. Create an empty Kotlin file in the same package as the test, and debug it in Android Studio by clicking the bug icon.


3. When the test enters an infinite loop, pause execution.


4. Debug further to identify the root cause of the issue.



If you have any question, let me know.

ashughes:
@takahirom Thanks for the suggestion. I'm stuck on step 2. I assume you mean to create an empty Kotlin file in the package `com.github.takahirom.roborazzi` as that is where the generated `RoborazziPreviewParameterizedTests` is. However, having an empty kotlin file there doesn't provide a way to breakpoint/debug the failing tests. Any idea what I'm missing?

Just to be clear, I don't have any tests written for roborazzi. I was just using the generated preview parameterized tests. Maybe I need to add a custom test in order to be able to breakpoint and figure our where things are getting stuck.

Also, the first 4 parameterized tests complete quickly (but fail with the `IllegalStateException` from this PR). It's on the 5th test that it starts taking minutes before failing with the same exception.

ashughes:
@takahirom I realized I could just copy/paste the generated `RoborazziPreviewParameterizedTests` into my own test file and I'm stepping through that.

It looks like the `View` returned by `ViewFinder.getView()` inside `ViewInteraction.doPerform()` has a 0 width/height and is being passed into `RoboComponent.View` inside `ImageCaptureViewAction.perform()`. This causes `View.fetchImage()` inside `ViewScreenshot.kt` to return `null`, which then throws the `IllegalStateException` since `rootComponent.image` is `null`.

Do you have any idea why the `View` has a 0 width/height? I'm not able to step through the espresso code as it says "Library source does not match the bytecode for class `ViewInteraction.java`.

takahirom:
Thank you. When I saw a view with a width and height of zero, it was not being attached to the Activity. However, I think this is not the case for your situation. Something like this.

```kotlin
ActivityScenario.launchActivity().onActivity{
  View(it).captureRoboImage()
}
```

takahirom:
However, I think you can check whether the View is attached to the Activity.

ashughes:
I see that `view.isAttachedToWindow = true` and `mRootView` is a `DecoreView[RoborazziTransparentActivity]`. I assume that means it is attached to the Activity.

takahirom:
@ashughes Thanks. If there is a setting that prevents us from taking screenshots, we need to add an error or warning in the Plugin. However, I'm not sure what happens in this case. Do you have any `robolectric.properties` settings or `unitTest` settings in `build.gradle`?

ashughes:
@takahirom I have:

```groovy
testOptions {
    unitTests {
        includeAndroidResources = true // Already had this before roborazzi
        all {
            systemProperties["robolectric.pixelCopyRenderMode"] = "hardware"
        }
    }
}
```

We already use robolectric for unit tests, but no special configurations other than via `@Config` annotations on test classes.

I cloned the roborazzi repo to see if I can repro the issue with the `sample-generate-preview-tests` module, but running `./gradlew :sample-generate-preview-tests:recordRoborazziDebug` passes.

I didn't realize the default output directory was in `build/outputs/roborazzi`. I checked my project and found that there in fact are some screenshots! So apparently only some of the tests are failing, not all of them. Looking at all the failed tests, I think the common denominator is that they all have dialogs.

ashughes:
The passing tests were still taking a long time to generate the screenshots. In fact, it seems to take an increasing amount of time. The first 4 take seconds, then the next take minutes, and just keeps increasing.

I happened to click on one of the passing tests and it contained error logs with `OutOfMemoryError`. I remembered seeing something about that in the docs, so I added `maxHeapSize = "4096m"` the setup in my last post. This appears to have fixed the issue as the tests now all take milliseconds to run. 🎉 

So it seems I've identified two issues:
1. Recording screenshots for composable previews with `Dialog` (and `AlertDialog`) results in the `IllegalStateException` added in this PR.
2. `OutOfMemoryError`s are not causing screenshot recording to fail, but instead to take a long time (I have not dug into this any further).

PR #134, opened 2023-08-03
Fix crash when the component size is zero and improve performance

takahirom:


roborazzi/src/main/java/com/github/takahirom/roborazzi/ComposeScreenshot.kt
```diff
@@ -26,12 +26,18 @@ fun SemanticsNode.fetchImage(recordOptions: RoborazziOptions.RecordOptions): Bit
...
 }
 
 
-fun Bitmap.crop(rect: Rect): Bitmap? {
-  val croppedBitmap = Bitmap.createBitmap(rect.width(), rect.height(), Bitmap.Config.ARGB_8888)
+fun Bitmap.crop(rect: Rect, recordOptions: RoborazziOptions.RecordOptions): Bitmap? {
+  if (rect.width() == 0 || rect.height() == 0) {
+    return null
+  }
+  if (rect.width() == this.width && rect.height() == this.height) {
```
takahirom, line 37: I may have to check if rect x y is zero

PR #151, opened 2023-09-06
[Proposal] :recycle: Added the ability to check the path of locally output images.

Corvus400:
# Motivation for issuing this PR.
I thought it would be useful to have the ability to see where the images are stored when responding to the following PRs.🙏
https://github.com/DroidKaigi/conference-app-2023/pull/1090


takahirom:
Thank you for the PR! If we add it here, we'll need to keep track of this variable in various places, making management challenging. I'm torn, but for now, I think the following approach might be better. What do you think?

Corvus400:
Sorry, I made a correction to the test before checking the comments, so I'll revise it again based on your suggestion!

Issue #153, opened 2023-09-16
[Seeking Ideas] Support for HTML Report Format


takahirom:
Currently, we don't offer support for HTML-based reports.  
I don't think we need an elaborate report file to start with.  
Here's a simple example:

```html
<table>
  <tr><td>File Name</td><td>Image</td></tr>
  <tr><td>fileName.png</td><td><img src='https://user-images.githubusercontent.com/1386930/246645185-579199d5-8e17-4f51-b990-de603ca36251.png' width='80%' /></td></tr>
</table>
```

<table>
  <tr><td>File Name</td><td>Image</td></tr>
  <tr><td>fileName.png</td><td><img src='https://user-images.githubusercontent.com/1386930/246645185-579199d5-8e17-4f51-b990-de603ca36251.png' width='80%' /></td></tr>
</table>

Any must-have features you'd like to see? Let me know!

jeprubio:
I think a report in html when running roborazzi would be great, it could use something like materializecss.com to look great and generate all the files (html + pngs) under build/reports.

If roboelectric is configured with a device (as: `@Config(qualifiers = RobolectricDeviceQualifiers.Pixel5)`) the resulting *compare.png image seem too big and I think they should be scaled down when shown in such an html. I would probably show the image scaled in the html and then in original size once an image is clicked.

takahirom:
> something like materializecss.com

Thanks for bringing this to my attention.
It seems that materializecss.com is abandonware, so considering a forked project might be a good idea.
https://materializeweb.com/

> I would probably show the image scaled in the html and then in original size once an image is clicked.

For now, I'm considering using width='80%' to appropriately resize the images.

takahirom:
I made a sample report file.
[index.html.zip](https://github.com/takahirom/roborazzi/files/12642146/index.html.zip)

takahirom:
> I think they should be scaled down when shown in such an html

I agree, and it's a great idea. However, I think it's better than nothing, so let's start by releasing it.

PR #155, opened 2023-09-17
Add HTML report

takahirom:
close https://github.com/takahirom/roborazzi/issues/153


include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResults.kt
```diff
@@ -19,6 +19,73 @@ data class CaptureResults(
...
+        append("<h3>$title (${images.size})</h3>")
+        val fileNameClass = "flow-text col s3"
+        val fileNameStyle = "word-wrap: break-word; word-break: break-all;"
+        val imgClass = "col s7"
+        val imgAttributes = "style=\"width: 100%; height: 100%; object-fit: cover;\""
+        append("<table class=\"highlight\">")
+        append("<thead>")
+        append("<tr class=\"row\">")
+        append("<td class=\"$fileNameClass\" style=\"$fileNameStyle\">File Name</td>")
+        append("<td class=\"$imgClass flow-text \">Image</td>")
```
jeprubio, line 45: It's not important, but perhaps using the `<th>` element, which defines a header cell in an HTML table, would be more suitable here
https://www.w3schools.com/tags/tag_th.asp

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/ResultSummary.kt
```diff
@@ -19,6 +19,32 @@ data class ResultSummary(
...
+    return """
+        <h3>Summary</h3>
+        <table class="highlight">
+            <thead>
+            <tr>
+                <td>Total</td>
+                <td>Recorded</td>
+                <td>Added</td>
+                <td>Changed</td>
+                <td>Unchanged</td>
```
jeprubio, line 32: These should probably also be `<th>`

jeprubio, line 32: These should probably also be `<th>` instead of `<td>`
```suggestion
                <th>Total</th>
                <th>Recorded</th>
                <th>Added</th>
                <th>Changed</th>
                <th>Unchanged</th>
```

PR #168, opened 2023-09-22
[Proposal] Add anchors to html report

jeprubio:
When looking at the html report there is this summary on top and then if you want to see the changed screenshots you have to manually scroll down. 
I thought it could be nice adding some html anchors so when any of these cells of recorded, added, changed or unchanged is clicked the browser goes to that section.

Roborazzi Anchors


takahirom (approved):
Looking great! It's definitely convenient!

PR #169, opened 2023-09-23
Show images in bottom sheet modal

jeprubio:
Once an image of the HTML report is clicked a bottom sheet modal is shown with that image.
In order to do that a imageBottomSheet is created in the html and then there is code in javascript that updates the modal content via javascript on image click.

Notice there is also a small unrelated change proposal of:
```
val imgAttributes = "style=\"width: 100%;...
```
for
```
val imgAttributes = "style=\"max-width: 100%;...
```
so that small images are not scaled up in the html report. If the original image was small (like a button) it was scaled up to match the width of the column table. No problem in reverting that but I think the report looks better if images are only scaled down or not scaled.

### DIFF
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResults.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResults.kt
@@ -52,22 +52,22 @@ data class CaptureResults(
           when (image) {
             is CaptureResult.Added -> {
               append("<td class=\"$fileNameClass\" style=\"$fileNameStyle\">${image.actualFile.name}</td>")
-              append("<td class=\"$imgClass\"><img $imgAttributes src=\"${image.actualFile.pathFrom(reportDirectoryPath)}\"/></td>")
+              append("<td class=\"$imgClass\"><img $imgAttributes class=\"modal-trigger\" src=\"${image.actualFile.pathFrom(reportDirectoryPath)}\" data-alt=\"${image.goldenFile.name}\"/></td>")
             }
 
             is CaptureResult.Changed -> {
               append("<td class=\"$fileNameClass\" style=\"$fileNameStyle\">${image.compareFile.name}</td>")
-              append("<td class=\"$imgClass\"><img $imgAttributes src=\"${image.compareFile.pathFrom(reportDirectoryPath)}\"/></td>")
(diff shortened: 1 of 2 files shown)


include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/RoborazziReportConst.kt
```diff
@@ -77,6 +92,22 @@ REPORT_TEMPLATE_BODY
 <script src="https://cdn.jsdelivr.net/npm/@materializecss/materialize/dist/js/materialize.min.js"></script>
 <script>
     M.AutoInit();
+    document.addEventListener('DOMContentLoaded', function() {
+        var modalInstance = M.Modal.init(document.getElementById('imageBottomSheet'), {});
+        var modal = document.getElementById('imageBottomSheet');
+        var modalImage = document.getElementById('modalImage');
+        var modalTriggers = document.querySelectorAll('.modal-trigger');
+        modalTriggers.forEach(function(trigger) {
+            trigger.addEventListener('click', function() {
```
takahirom, line 101: 👍

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResults.kt
```diff
@@ -52,22 +52,22 @@ data class CaptureResults(
           when (image) {
             is CaptureResult.Added -> {
               append("<td class=\"$fileNameClass\" style=\"$fileNameStyle\">${image.actualFile.name}</td>")
-              append("<td class=\"$imgClass\"><img $imgAttributes src=\"${image.actualFile.pathFrom(reportDirectoryPath)}\"/></td>")
+              append("<td class=\"$imgClass\"><img $imgAttributes class=\"modal-trigger\" src=\"${image.actualFile.pathFrom(reportDirectoryPath)}\" data-alt=\"${image.goldenFile.name}\"/></td>")
```
takahirom, line 55: How about adding class=\"modal-trigger\" to imgAttributes for consistency?

takahirom (approved):
I've added some comments, but if you're short on time, it's better to merge this change than to leave it.

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResults.kt
jeprubio: Yup, that makes a lot of sense [ce2037b](https://github.com/takahirom/roborazzi/pull/169/commits/ce2037bbcc8b88e541b99db80aab601cfab839b0)

PR #170, opened 2023-09-23
Generate Comparison Images Even When Initial Verification Fails

takahirom:
## Overview
This PR modifies the behavior of the verification task in Roborazzi. Previously, if the verification task failed, the process would stop, and no further comparison images would be generated. With this change, Roborazzi will continue to generate comparison images even if the initial verification fails.


takahirom:
copilot:all

timothyfroehlich:
So I've realized that when I was asking about this I was thinking narrowly about my current and basic use case where I am only capturing one image per test function. 

So three scenarios. I've asked about a subset of these, but I want to ask about all of them at once, just to be clear.

1: I have three `captureRoboImage()` calls within a test function and the first call fails image validation.
2: I have three test functions inside a test class and the first function fails image validation
3: I have three test classes and the first one fails image validation

In each scenario, with the current behavior, will I:

A: Get validation results for the later tests/calls?
B: Get image comparisons for the later test/calls?

takahirom:
Thanks for bringing this up. We might need to consider these specific test scenarios. Below is a summary of the current behavior:

| Scenario | Verification Behavior (`verifyRoborazziDebug`) | Comparison Behavior (`compareRoborazziDebug`)  |
|--------------------------|--------------------------------------|---------------------------|
| Multiple `captureRoboImage()` calls within a test function | If the first `captureRoboImage()` call throws an exception, subsequent calls won't be executed. However, you will still receive validation results and images from that initial call. | `captureRoboImage()` calls are expected to execute without issues, and you'll receive complete validation results and images, including those from the first call. |
| Multiple test functions within a test class | Even if the first test function throws an exception, the remaining functions will still be executed. You'll get validation results and images from all functions. | You'll have access to validation results and images from all functions. |

As for this PR, the behavior will be as follows:

| Scenario | Verification Behavior (`verifyRoborazziDebug`) | Comparison Behavior (`compareRoborazziDebug`)  |
|--------------------------|--------------------------------------|---------------------------|
| Multiple `captureRoboImage()` calls within a test function | The first `captureRoboImage()` call doesn't throw an exception, and the test function itself later throws an exception. You will receive all images. However, due to throwing the first `VerificationError` at the end of the test, you won't have access to additional validation results or stack traces from subsequent calls. | `captureRoboImage()` calls are expected to proceed as usual, and you'll receive complete validation results and images, including those from the first call. |

takahirom:
@timothyfroehlich 

Thanks for bringing this up. We might need to consider these specific test scenarios. Below is a summary of the current behavior:

| Scenario | Verification Behavior (`verifyRoborazziDebug`) | Comparison Behavior (`compareRoborazziDebug`)  |
|--------------------------|--------------------------------------|---------------------------|
| Multiple `captureRoboImage()` calls within a test function | If the first `captureRoboImage()` call throws an exception, subsequent calls won't be executed. However, you will still receive validation results and images from that initial call. | `captureRoboImage()` calls are expected to execute without issues, and you'll receive complete validation results and images, including those from the first call. |
| Multiple test functions within a test class | Even if the first test function throws an exception, the remaining functions will still be executed. You'll get validation results and images from all functions. | You'll have access to validation results and images from all functions. |

As for this PR, the behavior will be as follows:

| Scenario | Verification Behavior (`verifyRoborazziDebug`) | Comparison Behavior (`compareRoborazziDebug`)  |
|--------------------------|--------------------------------------|---------------------------|
| Multiple `captureRoboImage()` calls within a test function | The first `captureRoboImage()` call doesn't throw an exception, and the test function itself later throws an exception. You will receive all images. However, due to throwing the first `VerificationError` at the end of the test, you won't have access to additional validation results or stack traces from subsequent calls. | `captureRoboImage()` calls are expected to proceed as usual, and you'll receive complete validation results and images, including those from the first call. |

takahirom:
Incorporating this PR would bring some convenience, but it also presents a trade-off. Specifically, there is an existing best practice that suggests having one 'verify' operation per test. Adding this new functionality could complicate the structure of the test reports. Therefore, I'm unsure whether implementing this change is the right move.

takahirom:
Thank you. I won't be implementing these behavior changes, but I believe some refactoring can be done in this PR. Therefore, I'll modify this PR to focus solely on refactoring.

Issue #174, opened 2023-10-06
Flaky execution when a test task is run with `clean`


ZacSweers:
We run some of Gradle's cache validation experiments, and the one we created for tests flakes on roborazzi due to this: https://github.com/takahirom/roborazzi/blob/main/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt#L229

```
Caused by: kotlin.io.NoSuchFileException: /mnt/github/runner/_work/slack-android-ng/slack-android-ng/gradle-enterprise-gradle-build-validation/.data/01-validate-incremental-building/20231005T200748-651f7a04/build_project-to-validate/libraries/slack-kit/slack-kit/build/outputs/roborazzi: The source file doesn't exist.	
	at kotlin.io.FilesKt__UtilsKt.copyRecursively(Utils.kt:294)	
	at kotlin.io.FilesKt__UtilsKt.copyRecursively$default(Utils.kt:288)	
	at io.github.takahirom.roborazzi.RoborazziPlugin$apply$configureComponents$1$6$5.invoke(RoborazziPlugin.kt:220)	
	at io.github.takahirom.roborazzi.RoborazziPlugin$apply$configureComponents$1$6$5.invoke(RoborazziPlugin.kt:213)	
	at io.github.takahirom.roborazzi.RoborazziPlugin$apply$configureComponents$1$6.invoke$lambda$5(RoborazziPlugin.kt:213)
```

This happens sometimes when the command is `./gradlew clean :path:to:project:test`

takahirom:
Thank you for reporting this issue.

Regarding the NoSuchFileException error, we should be able to resolve this issue immediately by creating the folder. However, I am working on reproducing the error in the integration tests.

takahirom:
I have successfully reproduced the error and resolved it. Thank you for bringing this to our attention!

ZacSweers:
Thanks!

PR #175, opened 2023-10-06
Resolve the FileNotFoundException Issue with RoborazziPlugin

takahirom:
Fix https://github.com/takahirom/roborazzi/issues/174


include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProject.kt
```diff
@@ -222,6 +223,17 @@ dependencies {
 }
           """
       )
+      if (removeOutputDirBeforeTestTypeTask) {
```
takahirom, line 226: This should be After 👀

PR #199, opened 2023-10-30
Fix compatibility with Robolectric 4.11

takahirom:


takahirom:
Waiting the response from Robolectric
https://github.com/robolectric/robolectric/issues/8577

takahirom:
It looks like the fix will be released tommorow 👀

PR #205, opened 2023-11-05
Enhance Flexibility with Custom File Naming in Roborazzi via roboOutputName() Function

### LINKED ISSUE #202: Add a file name parameter to captureRoboImage
When roborazzi captures multiple images, it suffixes them with their indices: `test`, `test_1`, `test_2`, and so on. Using numbers for differentiating between screenshots isn't very useful. Can roborazzi add a parameter to `captureRoboImage()` (as an overload) to accept custom file names?

takahirom:
close #202


timothyfroehlich:
Love this change!

Issue #236, opened 2024-01-01
Migrating from the org.json Library to Gson


takahirom:
Why?
* Using the org.json library requires manual mapping for each class, which is time-consuming.
* Vulnerabilities in the org.json library appear frequently, making it difficult to maintain.
* Since AGP also utilizes Gson, consolidating to a single library helps avoid the introduction of numerous libraries.

sanao1006:
I would like to work on this issue!

- Migrating the `org.json` Library to `Gson`
- Pass all `CaptureResultTest` tests

Are there any other tasks that need to be completed in order to resolve this issue?
(E.g., pass tests for `sample-android`, etc.)

takahirom:
Thanks! I think that's all you have to do to solve this issue.

PR #238, opened 2024-01-03
Fix CaptureResultTest and Add plan test workflow

momomomo111:
I checked to see if the Json conversion test was working correctly to tackle Issue #236 and found that it was not working correctly.


takahirom:
Thanks. I didn't notice we broke the roborazzi-core tests. I added a test workflow to address this. 👀

PR #241, opened 2024-01-08
Use ComposeTestRule interface instead of concrete AndroidComposeTestRule class

GisoBartels:
This restores Roborazzi's compatibility with customized test rules.


GisoBartels:
Thanks a lot for the quick review.
Would it be possible to backport this change to the latest stable version?

takahirom:
@GisoBartels I don't consider it a serious bug, so would it be okay to include it in version 1.8.0-rc-2?

GisoBartels:
Yes, that's fine. Thanks.

takahirom:
Thank you for understanding. I'll make sure we can release it today.

takahirom:
@GisoBartels I've released it. Could you check it out?

GisoBartels:
Works flawlessly. Thanks for the quick release 👍

PR #248, opened 2024-01-19
refactor: migrating from the `org.json` Library to `Gson`

### LINKED ISSUE #236: Migrating from the org.json Library to Gson

sanao1006:
> - Using the org.json library requires manual mapping for each class, which is time-consuming.
> - Vulnerabilities in the org.json library appear frequently, making it difficult to maintain.
> - Since AGP also utilizes Gson, consolidating to a single library helps avoid the introduction of numerous libraries.

Migrate `org.json` to `Gson` for the above reasons

###  note
- Passed all `CaptureResultTest` tests

### related issue
Fixed #236

### DIFF
--- a/include-build/roborazzi-core/build.gradle
+++ b/include-build/roborazzi-core/build.gradle
@@ -61,6 +61,7 @@ kotlin {
         compileOnly libs.androidx.compose.ui.test.junit4
         api libs.androidx.test.espresso.core
         implementation libs.androidx.core.ktx
+        implementation libs.gson
       }
     }
 
--- a/include-build/roborazzi-core/build.gradle
+++ b/include-build/roborazzi-core/build.gradle
@@ -32,7 +32,7 @@ kotlin {
     }
     commonJvmMain {
       dependencies {
-        compileOnly libs.json
+        compileOnly libs.gson
         api libs.dropbox.differ
         implementation libs.junit
       }
@@ -46,7 +46,7 @@ kotlin {
     }
     jvmMain {
       dependencies {
-        implementation libs.json
+        implementation libs.gson
       }
     }
     jvmTest {
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
@@ -1,11 +1,12 @@
 package com.github.takahirom.roborazzi
 
+import com.google.gson.Gson
+import com.google.gson.JsonObject
 import java.io.File
 import java.io.FileReader
-import org.json.JSONObject
 
 sealed interface CaptureResult {
-  fun toJson(): JSONObject
+  fun toJson(): JsonObject
   val timestampNs: Long
   val compareFile: File?
   val actualFile: File?
@@ -20,11 +21,11 @@ sealed interface CaptureResult {
     override val compareFile: File?
       get() = null
 
-    override fun toJson(): JSONObject {
-      val json = JSONObject()
-      json.put("type", "recorded")
-      json.put("golden_file_path", goldenFile.absolutePath)
-      json.put("timestamp", timestampNs)
+    override fun toJson(): JsonObject {
+      val json = JsonObject()
+      json.addProperty("type", "recorded")
+      json.addProperty("golden_file_path", goldenFile.absolutePath)
+      json.addProperty("timestamp", timestampNs)
       return json
     }
   }
@@ -35,13 +36,13 @@ sealed interface CaptureResult {
     override val goldenFile: File,
     override val timestampNs: Long,
   ) : CaptureResult {
-    override fun toJson(): JSONObject {
-      val json = JSONObject()
-      json.put("type", "added")
-      json.put("compare_file_path", compareFile.absolutePath)
-      json.put("actual_file_path", actualFile.absolutePath)
-      json.put("golden_file_path", goldenFile.absolutePath)
-      json.put("timestamp", timestampNs)
+    override fun toJson(): JsonObject {
+      val json = JsonObject()
+      json.addProperty("type", "added")
+      json.addProperty("compare_file_path", compareFile.absolutePath)
+      json.addProperty("actual_file_path", actualFile.absolutePath)
+      json.addProperty("golden_file_path", goldenFile.absolutePath)
+      json.addProperty("timestamp", timestampNs)
       return json
     }
   }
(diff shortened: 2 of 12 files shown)


include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
```diff
@@ -20,11 +21,11 @@ sealed interface CaptureResult {
     override val compareFile: File?
       get() = null
 
-    override fun toJson(): JSONObject {
-      val json = JSONObject()
-      json.put("type", "recorded")
-      json.put("golden_file_path", goldenFile.absolutePath)
-      json.put("timestamp", timestampNs)
+    override fun toJson(): JsonObject {
+      val json = JsonObject()
```
takahirom, line 25: This is better than what we have now, but I though we could use auto mapping👀
https://github.com/google/gson/blob/main/UserGuide.md#object-examples

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
```diff
@@ -22,11 +23,11 @@ sealed interface CaptureResult {
...
 
     override fun toJson(): JsonObject {
-      val json = JsonObject()
-      json.addProperty("type", "recorded")
-      json.addProperty("golden_file_path", goldenFile.absolutePath)
-      json.addProperty("timestamp", timestampNs)
-      return json
+      val recorded = object {
+        val golden_file_path = goldenFile.absolutePath
+        val timestamp = timestampNs
```
sanao1006, line 28: Writing `Gson().toJson(this)` would result in the following Json string
```
{goldenFile: "...", timestampNs: ....}
```
This does not pass the test because of the key name.
So I created an object with the correct key name.
(Same reason for `Added`, `Changed` and `UnChanged`.)

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResults.kt
```diff
@@ -1,24 +1,21 @@
...
 ) {
   fun toJson(): JsonObject {
-    val json = JsonObject()
-    json.add("summary", summary.toJson())
-    val resultsArray = JsonArray()
-    captureResults.forEach { result ->
-      resultsArray.add(result.toJson())
+    val captureResultsObject = object {
+      val summary = resultSummary
+      val results = captureResults.map { it.toJson() }
```
sanao1006: Because a new object cannot be created if the variable named `summary` has a duplicate name, the name of the other one was modified.

sanao1006:
@takahirom 
Sorry, I have no idea what caused the CI failure.
Do you have any ideas?

include-build/roborazzi-core/build.gradle
```diff
@@ -46,7 +46,7 @@ kotlin {
     }
     jvmMain {
       dependencies {
-        implementation libs.json
+        implementation libs.gson
```
takahirom, line 49: We might need to add it to androidMain, as org.json is included in the Android Framework, whereas gson is not.

sanao1006: Ok. I'll try it!

sanao1006:
It appears that gson related errors are also occurring in diff test.
```
...
Caused by: com.google.gson.JsonIOException: Interfaces can't be instantiated! Register an InstanceCreator or a TypeAdapter for this type. Interface name: com.github.takahirom.roborazzi.CaptureResult
...
```
I'll try to investigate.

takahirom:
We might have to implement like this 👀 
https://github.com/google/gson/blob/main/gson/src/test/java/com/google/gson/functional/RuntimeTypeAdapterFactoryFunctionalTest.java#L69

takahirom:
@sanao1006 I have attempted to use Gson's automatic conversion, but it hasn't worked yet. Could you take a look at it?

sanao1006:
```
...
Caused by: java.lang.NullPointerException: Cannot invoke "com.google.gson.JsonElement.getAsString()" because the return value of "com.google.gson.JsonObject.get(String)" is null
	at com.github.takahirom.roborazzi.CaptureResult$JsonAdapter.deserialize(CaptureResult.kt:88)
	at com.github.takahirom.roborazzi.CaptureResult$JsonAdapter.deserialize(CaptureResult.kt:68)
...
```
From this error message, it seems that error handling is missing when this part is null
```kotlin
val type = json.asJsonObject.get("type").asString
```

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
```diff
@@ -1,124 +1,96 @@
...
+        is Added -> context.serialize(src, Added::class.java)
+      }
+    }
 
+    override fun deserialize(
+      json: com.google.gson.JsonElement,
+      typeOfT: java.lang.reflect.Type,
+      context: com.google.gson.JsonDeserializationContext
+    ): CaptureResult {
+      val type = json.asJsonObject.get("type").asString
```
sanao1006, line 88:
```
...
Caused by: java.lang.NullPointerException: Cannot invoke "com.google.gson.JsonElement.getAsString()" because the return value of "com.google.gson.JsonObject.get(String)" is null
	at com.github.takahirom.roborazzi.CaptureResult$JsonAdapter.deserialize(CaptureResult.kt:88)
	at com.github.takahirom.roborazzi.CaptureResult$JsonAdapter.deserialize(CaptureResult.kt:68)
...
```
From this error message, it seems that error handling is missing when this part is null
```kotlin
val type = json.asJsonObject.get("type").asString
```

sanao1006:
This is a bit of a departure from the purpose of this PR, but I found a typo. 
https://github.com/takahirom/roborazzi/blob/89c22868f27899299f33b1a1c99e89fb5208f944/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProject.kt#L314
Do you mind if I modify it within this PR?

sanao1006:
@takahirom 
Are there any other tasks still left to be resolved in this PR?

takahirom:
As you can see, the diff comments are not working. Maybe the branch name is the cause but I want to check if it works before marge.
I'm addressing this issue. But if you have any recommendations let me know.
https://github.com/takahirom/roborazzi/pull/253

PR #252, opened 2024-01-27
Check updating compose 1.6

takahirom:


jeprubio:
That small changes in the text dimensions could be due to includeFontPadding being set to false by default in compose 1.6.0 https://medium.com/androiddevelopers/fixing-font-padding-in-compose-text-768cd232425b

Issue #261, opened 2024-02-16
Roborazzi compare doesn't report changes when test task is taken from build cache


francescocervone:
# Short description
Basically when you have this case
```
> Task :app:testDebugUnitTest FROM-CACHE
> Task :app:finalizeTestRoborazziDebug
> Task :app:compareRoborazziDebug UP-TO-DATE
```
the compare doesn't work correctly.

# How to reproduce

I've created [a sample project](https://github.com/francescocervone/RoborazziCacheBug) to reproduce the issue.

1. Run `./gradlew :app:recordRoborazziDebug`
2. Change something in the composable (like adding an extra exclamation mark [here](https://github.com/francescocervone/RoborazziCacheBug/blob/main/app/src/main/java/com/example/roborazzicachebug/MainActivity.kt#L35))
3. Run `./gradlew :app:compareRoborazziDebug`
4. Remove the change done in step 2
5. Run `./gradlew :app:clean`
6. Run `./gradlew :app:recordRoborazziDebug` again
7. Do the same change done in step 2 (this is very important, because the tests must be cached)
8. Run `./gradlew :app:compareRoborazziDebug --console=plain` again
9. If `testDebugUnitTest` is taken `FROM-CACHE` then you are not going to see any `_compare.png` file n the roborazzi output folder.

The steps to reproduce aren't always working because for some reason that is not fully clear to me, `:app:testDebugUnitTest` is not taken from cache always, so you might need to retry the full flow a couple of times, but I think this is unrelated to Roborazzi.

Even though it looks a weird use case, this happens all the time in CI when you open a PR and add a commit after the first round of CI that is unrelated to the screenshot tests.
I'll make an example:
- You open a PR that changes some UI, hence some screenshots
- The Roborazzi compare reports the change in the first CI round
- Then you push a commit that changes some file in the `androidTest` source set (totally unrelated to the screenshot then)
- The CI runs again, Roborazzi compare included, but the `test` task gets retrieved from cache (as the production and test source set have not changed) and Roborazzi doesn't report the changes

takahirom:
Thanks! I'll take a look at this

takahirom:
Thanks to your detailed reproduction procedure, I was able to replicate the problem in a test, and I believe I have fixed it. Please wait for the release.
https://github.com/takahirom/roborazzi/pull/262/files#diff-4777a7ba6e88f41372fefd5df5b24eeff8f66045fe8ba0e886538bca112240d8R195

[...]

PR #262, opened 2024-02-17
Fix Roborazzi output cache

takahirom:
close https://github.com/takahirom/roborazzi/issues/261

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
@@ -14,14 +14,19 @@ import org.gradle.api.Task
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
+import org.gradle.api.provider.Provider
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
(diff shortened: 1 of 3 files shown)


include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -157,33 +174,37 @@ class RoborazziPlugin : Plugin<Project> {
...
             t.doLast {
-              t.infoln("Roborazzi: roborazziTestFinalizer.doLast")
-              // Copy all files from outputDir to intermediateDir
-              // so that we can use Gradle's output caching
-              t.infoln("Copy files from ${outputDir.get()} to ${intermediateDir.get()}")
-              // outputDir.get().asFileTree.forEach {
-              //   println("Copy file ${it.absolutePath} to ${intermediateDir.get()}")
-              // }
-              outputDir.get().asFile.mkdirs()
-              outputDir.get().asFile.copyRecursively(
```
takahirom, line 183: The problem of not being able to see the output directory occurs because the test task's outputDir is set to the intermediate directory (build/intermediates/roborazzi), which is used as the input directory for the test task (build/outputs/roborazzi). Therefore, when Gradle loads the cache, it only loads the intermediate directory from the cache. However, we did not copy the contents of the intermediate directory to the output directory. To resolve this, we now copy the intermediate directory to the output directory whenever the test task is skipped.

takahirom, line 183: The issue of not being able to access the output directory arises because we have configured build/intermediates/roborazzi as the output directory for the test task in order to use build/outputs/roborazzi as its input directory. This configuration is essential for calculating image differences. Therefore, when Gradle loads the cache, it only loads the intermediate directory from the cache. However, we did not copy the contents of the intermediate directory to the output directory. To resolve this, we now copy the intermediate directory to the output directory whenever the test task is skipped.

PR #273, opened 2024-02-29
Check robolectric.screenshot.hwrdr.native=true with 4.12-SNAPSHOT

takahirom:

PR #283, opened 2024-03-27
Hide action bar to avoid shadow creeping into images captured using captureRoboImage

lukas-mercari:
I tried out shadow rendering as outlined in #273 and noticed that there's a shadow from the top when using `captureRoboImage` with a composable parameter. This is visible in the diff for `manual_compose` in #273.
It seems that it's caused by the action bar of `RoborazziTransparentActivity` rendering right above the composable and thus "leaking" a shadow into the screenshot.

I tested the change by applying it on top of #273.


takahirom:
I didn't notice this. Looks great!

takahirom:
@lukas-mercari Thanks for your contribution!

PR #293, opened 2024-04-10
Enable screenshot recording by means of an extension property

realdadfish:
I have an Android Monorepo with about a dozen components and hundreds of modules, including many Robolectric UI tests. To avoid having to "enable" screenshot recording and report generation on a per `gradle.properties` basis I looked for alternatives to run the report generation of screenshots by default, but could not find any suitable. I make heavy usage of build convention plugins, so the easiest was to have a custom "my.plugin.test.screenshots" build convention that would apply your plugin and, at the same time, enable screenshot reporting by default via the plugin's extension for the respective test tasks.

I figured it makes no sense to have such a configuration for the verify and compare tasks; my primary use case really is to automatically _see_ what's wrong when a Robolectric UI test fails with some assertion, to aid debugging the problem. While one can omit the usage of the Gradle plugin altogether and just use the core library (together with a `System.setProperty("roborazzi.test.record", "enabled")` somewhere), I found it useful to re-use the HTML reporting capability that the plugin provides.

From what I see Roborazzi seems very feature-rich, eventually in the future things could get a bit more separated, so one thing does not require the existence of another.

## Release Notes

* **New Features**
  * Introduced `taskType` configuration option to specify Roborazzi modes (Record, Compare, Verify, VerifyAndRecord) directly in your Gradle build configuration files.

* **Documentation**
  * Updated build setup guide to document three activation paths for Roborazzi: Gradle properties defaults, taskType configuration, and direct task invocation with mode-specific overrides.

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
@@ -1,5 +1,7 @@
 package io.github.takahirom.roborazzi
 
+import com.github.takahirom.roborazzi.ExperimentalRoborazziApi
+import com.github.takahirom.roborazzi.RoborazziTaskType
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
@@ -656,6 +658,21 @@ class RoborazziGradleProjectTest {
       recordWithFilter2().output.run(::assertNotSkipped)
       checkRecordedFileExists("$screenshotAndName.testCapture1.png")
       checkRecordedFileExists("$screenshotAndName.testCapture2.png")
+
+  @Test
+  fun shouldNotRecordResultsByDefault() {
+    RoborazziGradleProject(testProjectDir).apply {
+      unitTest()
+      checkResultsSummaryFileNotExists()
+    }
+  }
+
+  @Test
+  fun shouldRecordResultsByDefaultIfExtensionIsConfigured() {
+    RoborazziGradleProject(testProjectDir).apply {
+      appBuildFile.taskType = RoborazziTaskType.Record
+      unitTest()
+      checkResultsSummaryFileExists()
     }
   }
 }
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
@@ -5,6 +5,7 @@ import com.github.takahirom.roborazzi.CaptureResults
 import com.github.takahirom.roborazzi.ExperimentalRoborazziApi
 import com.github.takahirom.roborazzi.InternalRoborazziApi
 import com.github.takahirom.roborazzi.RoborazziReportConst
+import com.github.takahirom.roborazzi.RoborazziTaskType
 import org.gradle.api.Action
 import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
@@ -47,6 +48,8 @@ private const val DEFAULT_TEMP_DIR = "intermediates/roborazzi"
 open class RoborazziExtension @Inject constructor(objects: ObjectFactory) {
   val outputDir: DirectoryProperty = objects.directoryProperty()
 
+  val taskType: Property<RoborazziTaskType> = objects.property(RoborazziTaskType::class.java)
+    .convention(RoborazziTaskType.None)
   @ExperimentalRoborazziApi
   val compare: RoborazziCompareExtension =
     objects.newInstance(RoborazziCompareExtension::class.java)
@@ -245,6 +248,7 @@ abstract class RoborazziPlugin : Plugin<Project> {
             isCompareRun.map { compareRunValue ->
               isRecordRunValue || isVerifyRunValue || isVerifyAndRecordRunValue || compareRunValue
                 || hasRoborazziTaskProperty(roborazziProperties)
+                || extension.taskType.orNull != RoborazziTaskType.None
             }
           }
         }
@@ -395,9 +399,10 @@ abstract class RoborazziPlugin : Plugin<Project> {
             val isTaskPresent =
               isAnyTaskRun(isRecordRun, isVerifyRun, isVerifyAndRecordRun, isCompareRun)
             // Task properties
-            if (!isTaskPresent) {
+            if (!isTaskPresent && extension.taskType.orNull == RoborazziTaskType.None) {
               test.systemProperties.putAll(roborazziProperties)
             } else {
+              val extensionTaskType = extension.taskType.orNull
               // Apply other roborazzi properties except for the ones that
               // start with "roborazzi.test"
               test.systemProperties.putAll(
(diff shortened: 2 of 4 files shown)


include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
```diff
@@ -451,4 +451,21 @@ class RoborazziGradleProjectTest {
...
+  }
+
+  @Test
+  fun shouldRecordResultsByDefaultIfExtensionIsConfigured() {
+    RoborazziGradleProject(testProjectDir).apply {
+      appBuildFile.recordEnabled = true
+      unitTest()
+      checkResultsSummaryFileExists()
+    }
+  }
```
takahirom, line 470: Thank you for adding tests👍

takahirom:
Thanks. I'm interested in understanding more about your use case for managing screenshots in Roborazzi tests. Specifically, in what situations would you prefer setting recordEnabled = false instead of not applying the plugin at all? Could you elaborate on the scenarios or reasoning behind this choice?

takahirom:
Please take a look at the test report

```kotlin
FAILURE: Build failed with an exception.

* Where:
Build file '/private/var/folders/n2/pt_35rc53tdgkld9531s2tfh0000gn/T/junit4772681132119589584/app/build.gradle.kts' line: 83

* What went wrong:
Script compilation error:

  Line 83: }roborazzi {
            ^ Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
                public fun Project.roborazzi(configure: Action<RoborazziExtension>): Unit defined in org.gradle.kotlin.dsl

1 error

* Try:
> Run with --debug option to get more log output.
> Run with --scan to get full insights.
```

realdadfish:
> Please take a look at the test report
> 
> ```kotlin
> FAILURE: Build failed with an exception.
> 
> * Where:
> Build file '/private/var/folders/n2/pt_35rc53tdgkld9531s2tfh0000gn/T/junit4772681132119589584/app/build.gradle.kts' line: 83
> 
> * What went wrong:
> Script compilation error:
> 
>   Line 83: }roborazzi {
>             ^ Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
>                 public fun Project.roborazzi(configure: Action<RoborazziExtension>): Unit defined in org.gradle.kotlin.dsl
> 
> 1 error
> 
> * Try:
> > Run with --debug option to get more log output.
> > Run with --scan to get full insights.
> ```

Ah, I was unable to run the tests locally, so I trusted on the CI, fixed the wrapping issue.

realdadfish:
> Thanks. I'm interested in understanding more about your use case for managing screenshots in Roborazzi tests. Specifically, in what situations would you prefer setting recordEnabled = false instead of not applying the plugin at all? Could you elaborate on the scenarios or reasoning behind this choice?

Well, this is a good question. IMHO whenever the plugin is applied it _should_ actually record screenshots by default, because this is the foundation of what the plugin always does, collect the individual screenshots that were taken in tests and create an overall report for it.

The reason why I introduced a boolean property here is mainly backwards compatibility. Users of the plugin are used to the fact that they _have_ to specifically tell the plugin to do _anything_, and I did not want to break that behavior. Also, since running these tasks in addition to what the test execution does there is surely a bit overhead that people eventually not want to pay each time. I'd expect that screenshot and non-screenshot tests live side-by-side in a module and most of the time the non-screenshot tests don't need any tasks to be run to collect data from them (or even let the screenshot code in the tests be executed).

realdadfish:
Ok, I figured I miss something still. Even though screenshot recording and report generation works with this change, during test execution this branch is executed anyways:

```kotlin
  if (taskType.isEnabled() && !roborazziSystemPropertyTaskType().isEnabled()) {
    println(
      "Roborazzi Warning:\n" +
        "You have specified '$taskType' without the necessary plugin configuration like roborazzi.test.record=true or ./gradlew recordRoborazziDebug.\n" +
        "This may complicate your screenshot testing process because the behavior is not changeable. And it doesn't allow Roborazzi plugin to generate test report.\n" +
        "Please ensure proper setup in gradle.properties or via Gradle tasks for optimal functionality."
    )
  }
```

I originally thought I would tackle this by this here https://github.com/takahirom/roborazzi/pull/293/files#diff-dc9dacb4aa08c2bfedbc7cbce3ce02898ab79e36f1dad6a7d5c314f819f3cb5bR322, however, that does not seem to work. Would you happen to know if I'm missing something else?

takahirom:
If users are looking to use Roborazzi for screenshot tests, they'll likely need to run the verify or compare tasks. As I understand it from this PR, setting recordEnabled = true means recording will always occur, which could be somewhat confusing. We should consider the options' priority and relevant use cases more carefully. I'll need some time to think over these changes.

realdadfish:
Hi @takahirom, have you thought about how to continue with this?

takahirom:
Thanks for the updates. While I appreciate the convenience this could provide, I believe we need to consider the overall API design more comprehensively.

Currently, we have several options for enabling recording:
1. `roborazzi.test.record=true` in `gradle.properties`
2. `recordRoborazziDebug` Gradle task
3. `RoborazziOptions(taskType = RoborazziTaskType.Record)` in code

This PR introduces `recordEnable = true` within the Roborazzi DSL, which diverges slightly by using the term "enable." This might cause confusion or overlap with the existing options, and it could potentially limit us to only "record" actions, which might not be clear.

Perhaps we could consider something like `roborazzi { taskType = RoborazziTaskType.Record }` to align with existing configurations. Additionally, establishing a clear priority order—`gradle.property < Gradle DSL < task < RoborazziOptions.taskType`—and documenting this could help clarify their usage and justify this addition.

realdadfish:
@takahirom I made some changes to my original PR - was it this what you imagined here? Where would be a proper place to document the behaviour?

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -338,9 +343,10 @@ abstract class RoborazziPlugin : Plugin<Project> {
             val isTaskPresent =
               isAnyTaskRun(isRecordRun, isVerifyRun, isVerifyAndRecordRun, isCompareRun)
             // Task properties
-            if (!isTaskPresent) {
+            if (!isTaskPresent && extension.taskType.orNull == RoborazziTaskType.None) {
               test.systemProperties.putAll(roborazziProperties)
             } else {
+              val taskType = extension.taskType.orNull
```
takahirom, line 349: How about using extensionTaskType as the variable name to clarify that this is the task type for the extension?

takahirom:
Sorry for being super late. I think it looks great, and I would like to merge this. However, I would like to have a test that checks the priority of the options with tasks and gradle.properties. For the documentation, you can add it here. Perhaps we can include this `<details><summary>Gradle extension setting for the default behavior of Roborazzi for the Gradle module</summary> The explanation would be here </details>` https://github.com/takahirom/roborazzi/blob/main/docs/topics/build_setup.md

<details><summary>Gradle extension setting for the default behavior of Roborazzi for the Gradle module</summary> The explanation would be here </details>

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
realdadfish: Good idea, will do.

realdadfish:
> Sorry for being super late. I think it looks great, and I would like to merge this. However, I would like to have a test that checks the priority of the options with tasks and gradle.properties. 

This is where I stumble - because frankly I didn't fully understood the current way it is designed right now. From what I read in the code there are warnings if the task is executed, but the property is not present (see `processOutputImageAndReport.kt:46ff`), if the current work mode is already that tasks override `gradle.properties` settings, this seems to be in conflict with this warning. I also don't see an existing test that would test tasks overriding `gradle.properties` settings, this is all a bit confusing :-/. 

I know my change adds even more complexity to an already complex implementation, and I start to feel bad about this.

> For the documentation, you can add it here. Perhaps we can include this `<details><summary>Gradle extension setting for the default behavior of Roborazzi for the Gradle module</summary> The explanation would be here </details>` https://github.com/takahirom/roborazzi/blob/main/docs/topics/build_setup.md

I added preliminary documentation, added it however to the table, and added a small preface that explains the concept behind.

realdadfish:
Hi @takahirom, any comment on my last comment? Thanks!

realdadfish:
Hi @takahirom, two years have passed. Any interest in getting this merged? If so I'd invest the time to bring this up-to-date with the most recent base version, if not I'd close the PR now. What do you think?

takahirom:
@coderabbitai Could you explain why we do this and what value we can get from implementing this while maintaining compatibility.

realdadfish:
@takahirom I updated the PR, let me know what you think. Thanks!

realdadfish:
There seems to be no interest to integrate this upstream. Would be cool if this would have been communicated more directly.

PR #361, opened 2024-05-16
Include Disabled semantic property in accessibility explanation

lukas-mercari:
This PR adds `Disabled: "true"` to the accessibility explanation when `SemanticsProperties.Disabled` is set (in Compose code).
I also included a test that includes UI elements for all the different semantic properties that are currently being printed.

### Background

We started using `Dump.AccessibilityExplanation` for a11y testing recently, and I noticed when working on Buttons that the output would be the same regardless of them being enabled or disabled. It made me wonder whether the `enabled` property was set correctly, so I thought it'd be good to reflect this in the test output.


takahirom (approved):
Looks great! Thank you for enhancing our accessibility features!

PR #363, opened 2024-05-17
[WIP] Cleanup intermediate dir usage, since there seem to be some problems restoring it when caches are used

lukas-mercari:
Thank you as always for your work and release 1.16.0! When there were no changes affecting screenshot tests, it reduced the CI/CD build time of the VRT task to about 30% of what it used to be. 👏 

I noticed one problem when I was checking with a local cache. This can be reproduced using the following commands:
```
./gradlew :sample-android:recordRoborazziDebug --build-cache
// Note: ./sample-android/build/outputs/roborazzi exists
rm -rf ./sample-android/build
./gradlew :sample-android:recordRoborazziDebug --build-cache // Cached build
// Note: Sometimes ./sample-android/build/outputs/roborazzi doesn't exist. 😱
./gradlew :sample-android:recordRoborazziDebug --build-cache // Read again from cache without deleting build dir.
// Note: ./sample-android/build/outputs/roborazzi exists!
```

It seems that there is some flakiness when restoring screenshots from the intermediate to the output dir. This was added in #128 (but it's not immediate clear to me why it was needed).

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProject.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProject.kt
@@ -107,11 +107,6 @@ class RoborazziGradleProject(val testProjectDir: TemporaryFolder) {
     File(testProjectDir.root, "app/$buildDirName/outputs/roborazzi").deleteRecursively()
   }
 
-  fun removeRoborazziAndIntermediateOutputDir() {
-    File(testProjectDir.root, "app/$buildDirName/outputs/roborazzi").deleteRecursively()
-    File(testProjectDir.root, "app/$buildDirName/intermediates/roborazzi").deleteRecursively()
-  }
-
   fun assertNotSkipped(output: String) {
     assert(output.contains("testDebugUnitTest' is not up-to-date because"))
   }
@@ -119,6 +114,10 @@ class RoborazziGradleProject(val testProjectDir: TemporaryFolder) {
 
   fun assertSkipped(output: String) {
     assert(output.contains("testDebugUnitTest UP-TO-DATE"))
+  }
+
+  fun assertFromCache(output: String) {
+    assert(output.contains("testDebugUnitTest FROM-CACHE"))
   }
 
   enum class BuildType {
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
@@ -59,25 +59,7 @@ class RoborazziGradleProjectTest {
       record()
       removeRoborazziOutputDir()
       val output = record().output
-      assertSkipped(output)
-
-      checkResultsSummaryFileExists()
-      checkRecordedFileExists("$screenshotAndName.testCapture.png")
-      checkRecordedFileNotExists("$screenshotAndName.testCapture_compare.png")
-      checkRecordedFileNotExists("$screenshotAndName.testCapture_actual.png")
-    }
-  }
-
-  @Test
-  fun recordWhenRemovedOutputAndIntermediate() {
-    RoborazziGradleProject(testProjectDir).apply {
-      removeRoborazziAndIntermediateOutputDir()
-      record()
-      removeRoborazziAndIntermediateOutputDir()
-      // should not be skipped even if tests and sources are not changed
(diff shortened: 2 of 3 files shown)


takahirom:
Thanks to the output from this PR, I discovered a bug that our integration tests had not covered. Therefore, I added tests and reverted the removal of test task inputs.
https://github.com/takahirom/roborazzi/pull/364

takahirom:
We might be able to modify the input dir for either recording or for comparing and verifying.

lukas-mercari:
@takahirom Nice, thanks for adding the test and reverting the change! Indeed, after looking at it again it seems that making the cache work isn't going to be as simple as we had hoped.
The other problem that I ran into (why I started to make changes) was that even with the caching enabled we relied on [copying files from the intermediate dir to the output dir](https://github.com/takahirom/roborazzi/blob/452696b61a25db6bef5f62c9c4742b1028493953/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt#L264). In my local runs, `isTestSkipped` was not always detected correctly, so sometimes the output dir ended up being empty. (Maybe due to the configuration cache?)

takahirom:
We might need to add additional logs for further investigation.

lukas-mercari:
Yes, I added some locally and it seemed that `TestTaskSkipEventsServiceProvider.expectingTestNames` wasn't initialised properly. (So, the skipped test run never gets added to `skippedTestTaskMap`.)

takahirom:
Thanks. I'm thinking of removing the TestTaskSkipEventsServiceProvider and isTestSkipped because consistency is more important. I also think the copied content isn't that concerning. What do you think about this?

```kotlin
              // if (isTestSkipped) {
              // If the test is skipped, we need to use cached files
              val startCopy = System.currentTimeMillis()
              intermediateDir.get().asFile.mkdirs()
              intermediateDir.get().asFile.copyRecursively(
                target = outputDir.get().asFile,
                overwrite = true
              )
              finalizeTestTask.infoln("Roborazzi: finalizeTestRoborazziTask Copy files from ${intermediateDir.get()} to ${outputDir.get()} end ${System.currentTimeMillis() - startCopy}ms")
              // }
```

takahirom:
I'll make a PR to check integration tests

lukas-mercari:
@takahirom Sorry for the late reply. I found some time to do a bit more research today.

#### `build/outputs/roborazzi` not being restored from cache

Since #364 I got the following error instead:
```
* What went wrong:
A problem was found with the configuration of task ':sample-android:testDebugUnitTest' (type 'AndroidUnitTest').
  - Type 'com.android.build.gradle.tasks.factory.AndroidUnitTest' property '$1' specifies directory '/Users/lukas/Dev/test/roborazzi/sample-android/build/outputs/roborazzi' which doesn't exist.

    Reason: An input file was expected to be present but it doesn't exist.

    Possible solutions:
      1. Make sure the directory exists before the task is called.
      2. Make sure that the task which produces the directory is declared as an input.

    For more information, please refer to https://docs.gradle.org/8.4/userguide/validation_problems.html#input_file_does_not_exist in the Gradle documentation.
```

It's a bit hard to reproduce, since I believe the configuration cache needs to kick in, but the steps are roughly:
```
// Make a change to the tests in `sample-android`.
./gradlew :sample-android:recordRoborazziDebug --build-cache
rm -rf ./sample-android/build
./gradlew :sample-android:recordRoborazziDebug --build-cache
rm -rf ./sample-android/build
./gradlew :sample-android:recordRoborazziDebug --build-cache
```

I think the error is essentially what is described in this issue: https://github.com/gradle/gradle/issues/2016
I made a fix [like this](https://github.com/lukas-mercari/roborazzi/commit/7e8d4f92dfffcf062accd487038624ab21d7cd44), and this brought back the problem with restoring the output directory instead.
Finally #366 fixes that issue.

#### Modifying the input dir depending on if we run in record or compare/verify mode

It seems that we cannot read `isRecordRun` in `configureEach`:
 ```
> Could not create task ':sample-android:testDebugUnitTest'.
 > Cannot query the value of this property because it has no value available`.
```

We could read `test.systemProperties["roborazzi.test.record"]` directly, but in that case the caching behaviour would only change when the user runs `testDebugUnitTest` directly (and not using `recordRoborazziDebug`).

I'm a bit out of ideas on how to approach this -- this seems to be a general limitation in the interaction between gradle and how our tasks are set up. (E.g. paparazzi [seems to have the problem reversed](https://www.github.com/cashapp/paparazzi/issues/1334#issuecomment-2027983201) since they [don't declare the output dir as an input](https://github.com/cashapp/paparazzi/blob/a205508811008e30994a0cf8f60b4692f9ea34cd/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L241).)

PR #365, opened 2024-05-17
Update toolchain to Java 17 for all modules

mannodermaus:
The source & target compatibility of all Roborazzi JAR/AAR artifacts remains at Java 11. This update to Java 17 could improve build speed a little, but primarily it paves the way towards matching the minimum Java version imposed by Android Gradle Plugin (since 8.0.0). Also, we will need to update to 17 eventually to address a failure we're seeing in [the `roborazzi-junit5` PR](https://github.com/takahirom/roborazzi/pull/355), which relies on an extension that was built against Java 17 internally.

Remove all usages of `JavaVersion.VERSION_11` and move them into the central `build.gradle` build script. Unfortunately, this still needs to be duplicated for the `include-build` projects, but at least those can use the same version definitions from the version catalog.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,10 +14,43 @@ plugins {
 }
 
 allprojects {
+  // Apply the same Java compatibility for each module
+  def javaVersion = libs.versions.javaTarget.get()
+  def toolchainVersion = libs.versions.javaToolchain.get()
+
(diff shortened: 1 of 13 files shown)


takahirom (approved):

PR #368, opened 2024-05-18
Migrate from java.io to kotlinx-io

### LINKED ISSUE #302: Migrate java.io.File Usage to Kotlin Multiplatform-Compatible Abstraction
We're using [java.io.File](https://github.com/search?q=repo%3Atakahirom%2Froborazzi%20java.io.File&type=code) which is not compatible with Kotlin Multiplatform (KMP). We need to replace it with a KMP-friendly alternative to achieve cross-platform compatibility like iOS

eyedol:
Fixes #302

Changes:

- [x] 1. Migrate these two classes `CaptureResult` and `CaptureResults` to use classes from `kotlinx.io` as initial implementation to evaluate if it can work. 
- [x] 2. Migrate the use of `CaptureResult` and `CaptureResults` across the project
- [x] 3. Migrate iOS implementations to make use of the migrated `CaptureResult` and `CaptureResults` classes.
- [x] 4. Migrate `RoboCanvas` to use classes from `kotlinx.io`.
- [ ] 5. Migrate `RoboCanvas` use of `kotlinx.io`. 
- [x] 6. Make sure all existing unit tests are passing as expected.
- [x] 7. Possible regression to make sure nothing is broken or introduced.

To make `RoboConvas` resuable interface across the different supported targets, the [issue with differ](https://github.com/dropbox/differ/pull/17) needs to be fixed so we can move `RoboCanvas` to `commonMain` for reuse.


eyedol:
@takahirom opened a PR to get something going. Currently using kotlinx.io lib. Seems straightforward though I'm hitting an issue with dropbox's differ library. It's not able to resolve iOS packages for some reasons. I saw the [PR](https://github.com/dropbox/differ/pull/16) you made to address this but it seems it didn't resolve it entirely as even using a snapshot release of the differ, I still faced the same issue.

Issue:

```kotlin
:roborazzi-compose-ios:iosArm64Main: Could not find com.dropbox.differ:differ-iosarm64:0.0.2.
Required by:
    project :roborazzi-compose-ios > com.dropbox.differ:differ:0.0.2
```

Might you know why?

takahirom:
@eyedol
I'm not sure either. So, I asked the maintainer.
https://github.com/dropbox/differ/pull/16#issuecomment-2119081492

include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/KotlinxIo.kt
```diff
@@ -0,0 +1,58 @@
...
+import kotlinx.io.readString
+import kotlinx.io.write
+
+val Path.absolutePath: String
+  get() = this.toString()
+
+val Path.nameWithoutExtension: String
+  get() = name.substringBeforeLast(".")
+
+fun Path.relativeTo(base: Path): Path {
```
takahirom, line 16: Looks great! When I tried this, I wasn't able to find a way to implement it. Could we add a small unit test for this?

eyedol: Good call. I'll add unit test for it.

eyedol: Added with 51e989a

eyedol:
> @eyedol
> I'm not sure either. So, I asked the maintainer.
> [dropbox/differ#16 (comment)](https://github.com/dropbox/differ/pull/16#issuecomment-2119081492)

@takahirom Opened a [PR on the differ](https://github.com/dropbox/differ/pull/17) project. Hopefully it fixes it

include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
```diff
@@ -14,19 +15,17 @@ import kotlinx.serialization.json.JsonDecoder
...
 @Serializable(with = CaptureResult.CaptureResultSerializer::class)
 sealed interface CaptureResult {
   val type: String
   val timestampNs: Long
-  val compareFile: File?
-  val actualFile: File?
-  val goldenFile: File?
+  val compareFile: Path?
+  val actualFile: Path?
+  val goldenFile: Path?
```
eyedol, line 25: @takahirom I was thinking of using `String` here to abstract away the `Path` class. That way we can decouple the use of the kotlinio package even further. Wdyt?

takahirom: As JSON is simply a string, I think it's appropriate to use a string here. I don't have a strong opinion about this.

takahirom (approved):
Looks great!

include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
eyedol: Thanks for your input. I ran away with your comment and went ahead and change it to `String` to abstract `kotlinx-io` class even away from the core code base

eyedol:
@takahirom should be ready for a merge. Will wait for the differ issue to be fixed then we can move `RoboCanvas` to common code.

takahirom:
@eyedol Thanks! Sorry for the delay—I was focused on stabilizing Roborazzi. I'll merge this change. Thanks for your contribution! It seems you're aware of what we need to do to migrate iOS to the existing implementation. Could you create some issues to address this?

eyedol:
Thanks for the merge. Sure. I made the ticket: https://github.com/takahirom/roborazzi/issues/383

takahirom:
@eyedol 
Thanks!

PR #374, opened 2024-05-20
Declare input directories as files to avoid issues when they don't exist

lukas-mercari:
I sometimes get this error:
```
* What went wrong:
A problem was found with the configuration of task ':sample-android:testDebugUnitTest' (type 'AndroidUnitTest').
  - Type 'com.android.build.gradle.tasks.factory.AndroidUnitTest' property '$1' specifies directory '/Users/lukas/Dev/test/roborazzi/sample-android/build/outputs/roborazzi' which doesn't exist.

    Reason: An input file was expected to be present but it doesn't exist.

    Possible solutions:
      1. Make sure the directory exists before the task is called.
      2. Make sure that the task which produces the directory is declared as an input.

    For more information, please refer to https://docs.gradle.org/8.4/userguide/validation_problems.html#input_file_does_not_exist in the Gradle documentation.
```

It's a bit hard to reproduce, since I believe the configuration cache needs to kick in, but the following steps usually work:
```
// Make a change to the tests in `sample-android`.
./gradlew :sample-android:recordRoborazziDebug --build-cache
rm -rf ./sample-android/build
./gradlew :sample-android:recordRoborazziDebug --build-cache
rm -rf ./sample-android/build
./gradlew :sample-android:recordRoborazziDebug --build-cache
```

I believe the issue is essentially what is described in https://github.com/gradle/gradle/issues/2016.

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
@@ -301,20 +301,20 @@ abstract class RoborazziPlugin : Plugin<Project> {
         .configureEach { test ->
           val resultsDir = resultDirFileProperty.get().asFile
           if (restoreOutputDirRoborazziTaskProvider.isPresent) {
-            test.inputs.dir(restoreOutputDirRoborazziTaskProvider.map {
+            test.inputs.files(restoreOutputDirRoborazziTaskProvider.map {
               if (!it.outputDir.get().asFile.exists()) {
                 it.outputDir.get().asFile.mkdirs()
               }
               test.infoln("Roborazzi: Set input dir ${it.outputDir.get()} to test task")
-              it.outputDir
+              it.outputDir.files()
             })
           } else {
-            test.inputs.dir(outputDir.map {
+            test.inputs.files(outputDir.map {
               if (!it.asFile.exists()) {
                 it.asFile.mkdirs()
               }
               test.infoln("Roborazzi: Set input dir $it to test task")
-              it
+              it.files()
             })
           }
           test.outputs.dir(intermediateDirForEachVariant.map {
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
@@ -306,7 +306,7 @@ abstract class RoborazziPlugin : Plugin<Project> {
                 it.outputDir.get().asFile.mkdirs()
               }
               test.infoln("Roborazzi: Set input dir ${it.outputDir.get()} to test task")
-              it.outputDir.files()
+              it.outputDir.files(".")
             })
           } else {
             test.inputs.files(outputDir.map {
@@ -314,7 +314,7 @@ abstract class RoborazziPlugin : Plugin<Project> {
                 it.asFile.mkdirs()
               }
               test.infoln("Roborazzi: Set input dir $it to test task")
-              it.files()
+              it.files(".")
             })
           }
           test.outputs.dir(intermediateDirForEachVariant.map {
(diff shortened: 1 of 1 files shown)


takahirom:
Thank you for your investigation. I believe the contents of the output directory are changed during the build process, causing the test to fail.

https://github.com/takahirom/roborazzi/blob/7e8d4f92dfffcf062accd487038624ab21d7cd44/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt#L109

>checkIfOutputIsUsed
java.lang.AssertionError: Expected count: 0, actual count: 1 summary:ResultSummary(total=1, recorded=0, added=0, changed=1, unchanged=0)
	at io.github.takahirom.roborazzi.RoborazziGradleProject.checkResultCount(RoborazziGradleProject.kt:327)
	at io.github.takahirom.roborazzi.RoborazziGradleProject.checkResultCount$default(RoborazziGradleProject.kt:312)
	at io.github.takahirom.roborazzi.RoborazziGradleProjectTest.checkIfOutputIsUsed(RoborazziGradleProjectTest.kt:204)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

https://github.com/takahirom/roborazzi/blob/main/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt#L204

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -301,20 +301,20 @@ abstract class RoborazziPlugin : Plugin<Project> {
...
           val resultsDir = resultDirFileProperty.get().asFile
           if (restoreOutputDirRoborazziTaskProvider.isPresent) {
-            test.inputs.dir(restoreOutputDirRoborazziTaskProvider.map {
+            test.inputs.files(restoreOutputDirRoborazziTaskProvider.map {
               if (!it.outputDir.get().asFile.exists()) {
                 it.outputDir.get().asFile.mkdirs()
               }
               test.infoln("Roborazzi: Set input dir ${it.outputDir.get()} to test task")
-              it.outputDir
+              it.outputDir.files(".")
```
takahirom, line 309: I was wondering if this implementation will encounter the same problem, but does it work?

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -301,20 +301,20 @@ abstract class RoborazziPlugin : Plugin<Project> {
         .configureEach { test ->
           val resultsDir = resultDirFileProperty.get().asFile
           if (restoreOutputDirRoborazziTaskProvider.isPresent) {
-            test.inputs.dir(restoreOutputDirRoborazziTaskProvider.map {
+            test.inputs.files(restoreOutputDirRoborazziTaskProvider.map {
```
takahirom, line 304: Could you please add a comment and a link here to explain why we use files instead of a directory?

lukas-mercari:
@takahirom Sorry, I ran out of time to finish this today. Unfortunately I don't have a good idea on how gradle works, so all changes take me a bit of time. 😄  Let me double check the behaviour in the next days, possibly also in combination with what we conclude in #363.

Judging from my local runs it works for resolving the error I mentioned in the description -- and thus I can reproduce the issue that #366 fixes again.

takahirom:
@lukas-mercari Thanks! Let me know if you think we should release these PRs.
https://github.com/takahirom/roborazzi/pull/374
https://github.com/takahirom/roborazzi/pull/366

takahirom:
@lukas-mercari 
If you feel like we're getting stuck, I think it's okay to release these changes as they are. I'd rather improve things gradually than aim for perfection. What do you think? If you need more time, that's fine too.

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
lukas-mercari: Done! Btw, I believe we can now clean up the following three lines that create the output dir if it doesn't exist -- let me do that in a follow up though.

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
lukas-mercari: I re-tested it locally, and it seems to work!

takahirom:
Thanks!

PR #379, opened 2024-05-24
Remove code that creates output dir if it doesn't exist

lukas-mercari:
Since the output dir is optional now, we can cleanup the code that generates it.


lukas-mercari:
`unitTestWhenRunTwice` is failing.
I think the reason is that during the first run, the output dir gets created, so the second run has different inputs. Before this PR the output dir was created during the configuration phase, so the second run would be skipped since an empty output dir existed both for the first and second runs.

takahirom:
I'm fine with merging this, but we must at least pass the test. I think it's great that we can set up by removing the output directory in the integration test before the build runs. However, I'm not sure if it's difficult.

PR #381, opened 2024-05-25
Use kotlinx-io 0.3.3 to avoid updating Kotlin

takahirom:


takahirom:
@eyedol I discovered that kotlinx-io 0.3.5 requires Kotlin 1.9.24, but we are currently using 1.9.21. Consequently, I've downgraded to kotlinx-io 0.3.3, which supports 1.9.22—still not ideal, but a better match for now. For reference, Kotlin 1.9.22 was released on December 21, 2023, and 1.9.24 on May 7, 2024.

PR #385, opened 2024-05-27
Fix issue where afterSuite runs for non-root suites

takahirom:


lukas-mercari (approved):
Should work (according to my local test)! (See [here](https://github.com/takahirom/roborazzi/pull/366#issuecomment-2133050591) for context on why it is needed.)

lukas-mercari:
Just wanted to confirm that this has resolved the issue. (No failures on CI/CD since I updated the version 12 hours ago.) 👏

PR #386, opened 2024-05-27
[WIP] Only use output dir as an input when roborazzi runs in compare or verify modes

lukas-mercari:
This is an attempt to make the roborazzi gradle tasks fully cacheable.

### Problem

Currently, Roborazzi's gradle plugin declares the output directory (where the golden screenshot files are stored) as an input. This is correct when the task runs in compare and verify modes, but it is not necessary when running record mode only.
The side-effect of this is that it is much more difficult to find an existing build cache entry.

### Solution

The basic idea is that we need to detect whether we run in verify/compare or record mode and then selectively set the input files. The complication is that (when running `recordRoborazzi*`) we cannot know whether we are recording, because the provider does not contain a value [at the time when inputs are configured](https://github.com/takahirom/roborazzi/blob/8839dc28e1d8ebc67d684fde4feacbbddae44a2d/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt#L264).

#### Attempt 1

Generally it seems that we can provide `properties` that are lazily evaluated. (Such as `isRecordRun`.) I attempted to create one (using `project.objects.directoryProperty()`), which I would then be able set similar to how `isRecordRun`'s value is assigned.

However, this was not successful:
```
> Cannot fingerprint input property : value 'file collection' cannot be serialized.
```
It seems that setting complex properties as input is unusual. See: https://discuss.gradle.org/t/task-with-a-complex-object-as-input/41385

#### Attempt 2

Instead, I thought that we don't actually need to provide the file collection -- but a hash of the file directory should be good enough to detect whether it has changed or not. Essentially I declared a property named `goldenFilesHash`, which I set to a hash of the output directory if roborazzi is running in verify/compare mode -- and otherwise to "".

### DIFF
--- a/include-build/roborazzi-gradle-plugin/build.gradle
+++ b/include-build/roborazzi-gradle-plugin/build.gradle
@@ -18,6 +18,7 @@ dependencies {
     // We don't use junit for plugin
     exclude group: 'junit', module: 'junit'
   }
+  implementation 'com.google.guava:guava:31.0.1-jre'
   implementation libs.kotlinx.serialization.json
   integrationTestDepImplementation libs.android.tools.build.gradle
   integrationTestDepImplementation libs.kotlin.gradle.plugin
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
@@ -59,7 +59,7 @@ class RoborazziGradleProjectTest {
       record()
       removeRoborazziOutputDir()
       val output = record().output
-      assertNotSkipped(output)
+      assertSkipped(output)
 
       checkResultsSummaryFileExists()
       checkRecordedFileExists("$screenshotAndName.testCapture.png")
@@ -74,10 +74,8 @@ class RoborazziGradleProjectTest {
       removeRoborazziAndIntermediateOutputDir()
       record()
       removeRoborazziAndIntermediateOutputDir()
-      // should not be skipped even if tests and sources are not changed
-      // when output directory is removed
       val output = record().output
-      assertNotSkipped(output)
+      assertFromCache(output)
 
       checkResultsSummaryFileExists()
       checkRecordedFileExists("$screenshotAndName.testCapture.png")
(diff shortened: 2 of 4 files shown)


include-build/roborazzi-gradle-plugin/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
     // We don't use junit for plugin
     exclude group: 'junit', module: 'junit'
   }
+  implementation 'com.google.guava:guava:31.0.1-jre'
```
lukas-mercari, line 21: Not sure if using guava is the best choice for generating file hashes. I roughly copied how to generate them from the gradle-checksum plugin: https://github.com/gradle/gradle-checksum/tree/master.

include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
```diff
@@ -89,7 +87,6 @@ class RoborazziGradleProjectTest {
   @Test
   fun recordWhenRunTwice() {
     RoborazziGradleProject(testProjectDir).apply {
-      record()
```
lukas-mercari, line 92: I removed this since the test is named "when run twice".

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -452,6 +451,14 @@ abstract class RoborazziPlugin : Plugin<Project> {
   private fun String.capitalizeUS() =
     replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.US) else it.toString() }
 
+  private fun FileCollection.generateHash(): String {
```
lukas-mercari, line 454: For the `sample-android` project this seems to perform well (takes 30-40ms to find the final hash) -- I haven't tried this with a module with a large amount of tests though.

lukas-mercari:
@takahirom I think I might have finally found a solution for the gradle cache issue. It's not super straightforward unfortunately, but I'd be happy for hear any feedback!

takahirom:
Using hashes is one option, but I thought we could avoid using hashes for the record task by using empty files. So, I made a PR for checking
https://github.com/takahirom/roborazzi/pull/389

takahirom:
@lukas-mercari I think Gradle might be calculating the hash as well, and I believe it is better if we can rely on Gradle for this. So what do you think of https://github.com/takahirom/roborazzi/pull/389 ?

takahirom:
I'll close this PR as the fix is released!

Issue #176, opened 2023-10-08
Robolectric Emulator(Roborazzi Emulator)


takahirom:
I've realized that we can create an emulator using Robolectric Native Graphics. This is just a demo, but it works. I'm wondering if there are any use cases for this.

https://github.com/takahirom/roborazzi/assets/1386930/2b0560ff-b361-4610-8ea8-d753f487edfd

JoseAlcerreca:
I want to know more :) I guess it takes a screenshot on idle and displays it on the UI?

jeprubio:
Could that be useful to run the tests on multiple devices/configurations? 

Something like what [Test Harness](https://google.github.io/accompanist/testharness/) does for compose.

I mean without having to create multiple classes with annotations like `@Config(qualifiers = RobolectricDeviceQualifiers.*)`, which is the only way I can think of to achieve that for now.

sergio-sastre:
Probably for debugging purposes if you can see what is displayed before the screenshot happens, and you give an option to “stop/sleep” for x seconds

takahirom:
Thank you for your feedback. I think it will be useful for debugging. Here's what I'm thinking:
The reason I use a function is that we might want to debug a specific function rather than running all tests.

```kotlin
  @Test
  fun emulator() {
    onView(ViewMatchers.isRoot())
      .roboNativeGraphicsEmulator()
  }
```

<img width="1369" alt="image" src="https://github.com/takahirom/roborazzi/assets/1386930/2eb0fc68-f7c4-4ee2-bff4-b57b696a0ca5">

sergio-sastre:
> Thank you for your feedback. I think it will be useful for debugging. Here's what I'm thinking: The reason I use a function is that we might want to debug a specific function rather than running all tests.
> 
> ```kotlin
>   @Test
>   fun emulator() {
>     onView(ViewMatchers.isRoot())
>       .roboNativeGraphicsEmulator()
>   }
> ```
> 
> (when you click `Dump`) <img alt="image" width="1369" src="https://user-images.githubusercontent.com/1386930/274309571-2eb0fc68-f7c4-4ee2-bff4-b57b696a0ca5.png">

Thinking carefully, this makes sense not only for Roborazzi tests but also for Robolectric UI tests with RNG.

Sometimes they fail and I resort to run them on the emulator (shared test pattern) to see why they fail. This is definitely sth that would help debugging such tests without the need to making them “shared” between instrumentation & unit tests

[...]

PR #387, opened 2024-05-27
POC - RNG Emulator

yschimke:
ClientMain (Client -> gRPC -> Server) spits out a similar PNG for SimplePreview, that running on an emulator does. 

Quick and dirty POC for an alternative approach to https://github.com/takahirom/roborazzi/issues/176

- Runs the emulator controller gRPC that is used by Android Studio as well as https://github.com/google/android-emulator-webrtc
- Implements a few methods of that controller getGps, setGps, getBattery, getScreenshot
- Implements a custom gRPC service for launching a Compose preview
- Server and Client running separately communicating with Square Wire gRPC

Current Hacks

- The server should run as a typical Main method, instead of piggybacking on Robolectric Junit runners, but this was simpler for constructing
- Should maintain different Robolectric sandboxes concurrently 
- Should support client providing the classpath with current classes, instead of limited to current project
 
 Todo (after some discussion)
 - Cleanup and comment code
 - demonstrate it working with correct threading and a main with different sandboxes
 - implement enough to get the web rtc running and events
 - demonstrate running as a VS code preview?


yschimke:
a.png 
image

emulator 
image

yschimke:
For discussion only at the moment, happy to close, but wanted to see if it's viable.

takahirom:
@yschimke 
Looks amazing.
How can we try this? I found some main methods and test code.

Maybe like this?

```
./gradlew emulator-server:testDebugUnitTest
```

And then run the `emulator-client` test main() in Android Studio?

yschimke:
At the moment it's clunky. I'm running the server  in junit temporarily. It saves implementing all the sandbox config logic for a POC.

So run the testServer unit test method, it will print out it's started on port 8080.

Then run ClientMain. I had to turn off building the project on launch otherwise it waits for the unit test to finish which never does.

Sorry it's in such a rough state, but wanted to discuss the approach first

takahirom:
I believe the main point would be to provide an interface shared with the Android Emulator. It sounds very promising. I tried using the android-emulator-webrtc, but setting it up with SSL and CORS is a bit difficult. It would be great if we could avoid these issues for users.

yschimke:
@takahirom what I might try is a minimal project, outside roborazzi. One composable, one preview. Implement there and include a web page, showing android-emulator-webrtc.

This is probably a slow burner project, but wanted to see if it aligns with what you were envisioning.

takahirom:
Thanks. I think there are two use cases for the RNG emulator:

* When we run instrumentation tests, we can observe the emulator's behavior. This approach is very intuitive for debugging the tests.
* I would like to debug the app's implementation, similar to how we do with a standard emulator.

Sharing an interface with the emulator isn't necessary, but it would be intuitive to use the emulator in the usual way for debugging. There may be some challenges, however. If this becomes a blocker, we can proceed without the interface.

PR #389, opened 2024-05-28
Use empty files for record task

takahirom:
The challenges we aim to address

https://github.com/takahirom/roborazzi/pull/386

> Problem
Currently, Roborazzi's gradle plugin declares the output directory (where the golden screenshot files are stored) as an input. This is correct when the task runs in compare and verify modes, but it is not necessary when running record mode only.
The side-effect of this is that it is much more difficult to find an existing build cache entry.


include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -261,27 +272,35 @@ abstract class RoborazziPlugin : Plugin<Project> {
...
-                it.outputDir.get().asFile.mkdirs()
-              }
-              test.infoln("Roborazzi: Set input dir ${it.outputDir.get()} to test task")
-              it.outputDir.files(".")
-            })
-          } else {
-            test.inputs.files(outputDir.map {
-              if (!it.asFile.exists()) {
-                it.asFile.mkdirs()
+          test.inputs.files(
```
lukas-mercari, line 275: I didn't know we can pass a `Provider` to `files`! (In fact, this is roughly what I was trying to do originally, but instead I was calling `isRecordRun.get()` which fails.) 👍

lukas-mercari (approved):
I think the integration tests are sufficient to prove that it works, but I doubled checked locally and it seems to work well! 👍 
Thanks a lot, this is indeed cleaner than my attempt in #386!

PR #400, opened 2024-06-07
Add rtl sample

takahirom:


yschimke:
Is this enough to test RTL? Don't those hardcoded strings need to be resources?

PR #404, opened 2024-06-10
Read external resources locally when generating html report

### LINKED ISSUE #295: HTML report should inline external resources
Currently, the generated HTML report links material.css and material.js externally. This is an issue in CI systems like Jenkins, where people like me eventually want to publish the report to (through the [HTML Publisher Plugin](https://plugins.jenkins.io/htmlpublisher/) for example), because there the report is loaded inside an `<iframe>` and [CSP rules apply](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). To then let the report render properly, one would have to white-list domains like `cdn.jsdelivr.net`, which is not only a burden to do (and not even possible in "hosted" environments), but also a big security hole, because when this domain is whitelisted, basically any NPM script is able to be loaded.

So I'd heavily vote for just inlining the needed CSS / JS during the build process and not link external resources in the report, at all.

eyedol:
Fixes #295 

Changes:

- Use [webjars](https://www.webjars.org/) to add the `materialize.min.css`, `materialize.min.js`, `material-icons` resources to the generated html report.
- Move the reports html from a string constant to a `index.html` file for easier editing
- Reference all the external resources as a relative local path in `index.html`
- Generate the html report based on the local `index.html` file


include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -255,12 +256,15 @@ abstract class RoborazziPlugin : Plugin<Project> {
...
               reportFile.parentFile.mkdirs()
+              writeAssetsToRoborazziReportsDir(reportFile.parentFile)
+              val reportHtml = readIndexHtmlFile()
               reportFile.writeText(
-                RoborazziReportConst.reportHtml.replace(
+                reportHtml?.replace(
                   oldValue = "REPORT_TEMPLATE_BODY",
                   newValue = roborazziResults.toHtml(reportFile.parentFile.absolutePath)
-                )
+                ) ?: ""
```
eyedol, line 267: Perhaps error in this case?

takahirom:
Thanks! The code looks good. However, I'm concerned about the LICENSE. I'm not sure if it is appropriate to include these files. I'm really sorry for noticing this late. 🙇

eyedol:
> Thanks! The code looks good. However, I'm concerned about the LICENSE. I'm not sure if it is appropriate to include these files. I'm really sorry for noticing this late. 🙇

That's a good concern to have. Checking the license for `materialize.min.css`, `materialize.min.js` it's the MIT license so we should be good. I'm no lawyer though, if you are able to verify with a lawyer it would be great. But personally I think it's fine as we're respecting everything stated in the MIT license.

> We have made these icons available for you to incorporate them into your products under the [Apache License Version 2.0](http://www.apache.org/licenses/LICENSE-2.0.txt). Feel free to remix and re-share these icons and documentation in your products. We'd love attribution in your app's about screen, but it's not required.

For the material icons it's Apache 2 license [1]. I believe we're good with this too.

[1]: https://developers.google.com/fonts/docs/material_icons#licensing

takahirom:
I agree with you. However, we might need to include both the Apache 2.0 and MIT licenses in our documentation. This could make it look a bit strange to users.

eyedol:
Oh I see. Is this to credit them? The license info are clear in the header of the minified files in case you're not aware.

takahirom:
Well, it is tough, but if we separate the repository into something like html-report-resources and make sure the license for Materialize is MIT and for the font is Apache 2.0, it might work. Maven includes license information, so this could be a viable solution.

eyedol:
You suggesting we create an entire separate repo to host the, the css and js files or create a new folder(to host the external files) in the same repo as this?

takahirom:
I found something 👀 Could we use this..?
https://mvnrepository.com/artifact/org.webjars/materializecss/1.0.0

eyedol:
That's gold. That should work. I'll look into it then. Thanks

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
```diff
@@ -0,0 +1,78 @@
+package io.github.takahirom.roborazzi
+
+import org.webjars.WebJarVersionLocator
+import java.io.File
+
+class WebAssets private constructor(private val webJarVersionLocator: WebJarVersionLocator) {
```
takahirom, line 6: 👍

eyedol: Thanks. The CSS from the webjar packing seems a little old. The styling is a little off from `main` Looking to see if there is a recently one published with webjar.

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -32,6 +32,8 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithTests
 import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
 import org.jetbrains.kotlin.gradle.targets.native.KotlinNativeBinaryTestRun
 import org.jetbrains.kotlin.gradle.targets.native.tasks.KotlinNativeTest
+import org.webjars.WebJarVersionLocator
```
takahirom, line 35:
```suggestion
```

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -50,6 +52,9 @@ open class RoborazziExtension @Inject constructor(objects: ObjectFactory) {
 @Suppress("unused")
 // From Paparazzi: https://github.com/cashapp/paparazzi/blob/a76702744a7f380480f323ffda124e845f2733aa/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
 abstract class RoborazziPlugin : Plugin<Project> {
+
+  private val webJarVersionLocator = WebJarVersionLocator()
+
```
takahirom, line 57:
```suggestion
```

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
takahirom: I noticed it, but it's a subtle change. So I think it is ok

takahirom (approved):
Looks great! Thank you for your patience in fixing this!

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -50,6 +52,9 @@ open class RoborazziExtension @Inject constructor(objects: ObjectFactory) {
 @Suppress("unused")
 // From Paparazzi: https://github.com/cashapp/paparazzi/blob/a76702744a7f380480f323ffda124e845f2733aa/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
 abstract class RoborazziPlugin : Plugin<Project> {
+
+  private val webJarVersionLocator = WebJarVersionLocator()
```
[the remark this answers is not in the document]
eyedol: Yes. It can be removed now. Thanks for catching it

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
eyedol: Yeah, however not all the tabs shows. That I think needs to be fixed.

eyedol: Resolved it with [9864d9d](https://github.com/takahirom/roborazzi/pull/404/commits/9864d9d2754de16e8fd7d7ae10439c62088048d3). All the tabs shows now though we lost the fixed sized tabs. I believe it should be okay?

takahirom:
I'll merge this PR. Since I don't want to include the JS files in the repository history, I'll squash and merge. 

PR #412, opened 2024-06-20
Add unit test for WebAssets class

eyedol:
Follow up to #404.

Add missing unit test for `WebAssets` class

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
@@ -64,7 +64,7 @@ class WebAssets private constructor(private val webJarVersionLocator: WebJarVers
 
   companion object {
 
-   internal val assetPathsMap = mapOf(
+   private val assetPathsMap = mapOf(
       "materializecss" to listOf(
         "css/materialize.min.css",
         "js/materialize.min.js",
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
@@ -4,9 +4,11 @@ import org.webjars.WebJarVersionLocator
 import java.io.File
 
 class WebAssets private constructor(private val webJarVersionLocator: WebJarVersionLocator) {
-  private val materializeCss = "materializecss"
-  private val materialIcons = "material-design-icons"
   private val webJarResource = "resources"
+
+  val assets = assetPathsMap.entries.flatMap {
+    assets -> assets.value.map { it.substringAfterLast("/") }
+  }
 
   fun writeToRoborazziReportsDir(reportDir: File) {
     writeLocalAssetsToRoborazziReportsDir(reportDir)
@@ -22,16 +24,7 @@ class WebAssets private constructor(private val webJarVersionLocator: WebJarVers
   }
 
   private fun writeWebJarAssetsToRoborazziReportsDir(reportDir: File) {
-    mapOf(
-      materializeCss to listOf(
-        "css/materialize.min.css",
-        "js/materialize.min.js",
-      ),
-      materialIcons to listOf(
-        "material-icons.css",
-        "MaterialIcons-Regular.ttf",
-      )
-    ).forEach { (key, value) ->
+    assetPathsMap.forEach { (key, value) ->
       value.forEach { exactPath ->
         webJarVersionLocator.locate(key, exactPath)?.let {
           writeAssets(
@@ -70,6 +63,17 @@ class WebAssets private constructor(private val webJarVersionLocator: WebJarVers
   }
 
   companion object {
+
+    val assetPathsMap = mapOf(
+      "materializecss" to listOf(
+        "css/materialize.min.css",
+        "js/materialize.min.js",
+      ),
+      "material-design-icons" to listOf(
+        "material-icons.css",
+        "MaterialIcons-Regular.ttf",
+      )
+    )
 
     fun create(
       webJarVersionLocator: WebJarVersionLocator = WebJarVersionLocator()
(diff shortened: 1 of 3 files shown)


include-build/roborazzi-gradle-plugin/src/test/java/io/github/takahirom/roborazzi/WebAssetsTest.kt
```diff
@@ -0,0 +1,31 @@
...
+class WebAssetsTest {
+
+  @get:Rule
+  val tmpDir: TemporaryFolder = TemporaryFolder.builder().assureDeletion().build()
+
+  @Test
+  fun testWritesWebAssetsSuccessfullyToReportsDirectory() {
+    val reportDir = tmpDir.newFolder("roborazzi-reports")
+    val cssFile = File(reportDir, "assets/report-style.css")
+    val materializeCssFile = File(reportDir, "assets/materialize.min.css")
```
takahirom, line 18: Great!
Could we have a shared file list in WebAssets? I would like to make it easier to update the assets library.

```
val files = assetsPaths.entiries.flatMap{ it.values.map { it.substringAfterLast("/") } }
```

https://github.com/takahirom/roborazzi/blob/ffa622267de2bdba81d1a2222f7f8f7cee5db9e8/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt#L27

```
companion object {
  val assetsPaths = mapOf(
      materializeCss to listOf(
        "css/materialize.min.css",
        "js/materialize.min.js",
      ),
      materialIcons to listOf(
        "material-icons.css",
        "MaterialIcons-Regular.ttf",
      )
    )
}
```

eyedol: Yes we can do that. You want assetsPaths exposed publicly too?

takahirom: Can we use internal or @ VisibleForTesting? But I'm not sure if we can use @ VisibleForTesting in JVM. 👀

eyedol: Oh if it's for use with test, I prefer to keep it private then reconstruct it by hand in test since the values are hardcoded, it can be caught in unit test in case someone changes it in the SUT.

eyedol: I believe we can use `@VisibleForTesting` in jvm tests

eyedol: > Oh if it's for use with test, I prefer to keep it private then reconstruct it by hand in test since the values are hardcoded, it can be caught in unit test in case someone changes it in the SUT.

@takahirom I was hoping to hear your thoughts on this. Should I make it `internal` or `private`?

takahirom: @eyedol 
Sorry for being late. I see your point about tests breaking when the file list changes. While it's not necessarily a bad thing, it does make maintenance a bit challenging. Here are three things I'd like to check in the test:

1. The files we want to include should be in the report folder to verify that index.html can use the files through the shared file list in the SUT.
2. The files should not be empty to confirm that we have the files correctly.
3. The report index.html should include the files to ensure that index.html is using the files.

eyedol: Sorry for the late response. I understand where you coming from though I think in this case it's fine to cause a duplicate as it could potentially catch an issue but yeah, I'll make it internal for reuse then.

eyedol: BTW, I changed the visibility to internal with 3e179ab

eyedol: @takahirom if all is good could you merge when you get the chance otherwise happy to fix anything I missed. Thanks

takahirom: Sorry for being late. What I wanted is to automate this using WebAssets.assetPathsMap so that we don't have to modify tests whenever files are changed.


```
    assertTrue(cssFile.exists())
    assertTrue(materializeCssFile.exists())
    assertTrue(materializeJsFile.exists())
    assertTrue(materialIconsCssFile.exists())
    assertTrue(materialIconsTtfFile.exists())
```

↓

```
val assetFiles = assetsPaths.entiries.flatMap{ it.values }.map { File(it) }

WebAssets.create().writeToRoborazziReportsDir(reportDir)

assetFiles.forEach { file->
  assertTrue(file.exists())
}
```

eyedol: Done with 04b0051

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
```diff
@@ -71,6 +64,17 @@ class WebAssets private constructor(private val webJarVersionLocator: WebJarVers
 
   companion object {
 
+   internal val assetPathsMap = mapOf(
```
takahirom, line 67: If we don't use this in tests or elsewhere, could we make this private?

```suggestion
   private val assetPathsMap = mapOf(
```

takahirom (approved):
Communicating with text can be challenging, but thank you for your patience in adding this!

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/WebAssets.kt
eyedol: oh, thought you'll need it else where for use. Thanks for the fix and making it `private`

eyedol:
> Communicating with text can be challenging, but thank you for your patience in adding this!

No problem. Thanks for the merge 🙏🏿

PR #429, opened 2024-07-11
Generate Roborazzi images via its IntelliJ Plugin

### LINKED ISSUE #424: Generate Images by Invoking Gradle Tasks Using the Roborazzi IntelliJ Plugin
I think it would be great to have a button to generate and refresh images in the plugin.

We can get Gradle module information and tasks.

```kotlin
plugins.set(listOf(
    "java",
    "Kotlin",
    "com.intellij.gradle", // add this
  ))
```

```kotlin
  private fun setUpGradleData(psiFile: KtFile) {
    val module = psiFile.module
    if (module == null) {
      println("module is null")
      return
    }
    val findGradleModuleData = GradleUtil.findGradleModuleData(module)
    println("findGradleModuleData:" + findGradleModuleData)
    findGradleModuleData?.let {
      println("findGradleModuleData:${it.data}")
      println("it.getDirectoryToRunTask():" + it.data.getDirectoryToRunTask())
      val taskNodeData = ExternalSystemApiUtil.findAll(it, ProjectKeys.TASK)
      taskNodeData.forEach {
        println("taskNodeData:" + it)
      }
    }
  }
```

> ./gradlew :roborazzi-idea-plugin:runIde

```
virtualFilePath/.../git/roborazzi/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
findGradleModuleData:ModuleData: :sample-android
findGradleModuleData::sample-android
it.getDirectoryToRunTask():/.../git/roborazzi/sample-android
taskNodeData:TaskData: projects
taskNodeData:TaskData: lintReportRelease
taskNodeData:TaskData: packageReleaseResources
taskNodeData:TaskData: restoreOutputDirRoborazzi
taskNodeData:TaskData: bundleLibResDebug
```

And it seems that we can run the task using the plugin:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/4414033579922-Run-Gradle-task-from-plugin-action

I think we can filter the tasks by recordRoborazzi, compareRoborazzi, and verifyRoborazzi, and show a drop-down menu to allow the user to select and run the task.

eyedol:
Fixes #424 

- Add a dropdown menu to a toolbar for selecting task to run

Initial implementation to solicit early feedback...

### DIFF
--- a/roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/PreviewViewModel.kt
+++ b/roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/PreviewViewModel.kt
@@ -1,0 +1,255 @@
+package com.github.takahirom.roborazzi.idea.preview
+
+import com.github.takahirom.roborazzi.idea.settings.AppSettingsState
+import com.intellij.openapi.fileEditor.FileEditorManager
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.roots.ProjectRootManager
+import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.impl.source.tree.CompositeElement
+import com.intellij.psi.util.PsiTreeUtil
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import kotlinx.coroutines.yield
+import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtDeclaration
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtFunction
+import java.io.File
+
+class PreviewViewModel {
+
+  var coroutineScope = MainScope()
+  val imagesStateFlow = MutableStateFlow<List<Pair<String, Long>>>(listOf())
+  private val lastEditingFileName = MutableStateFlow<String?>(null)
+  val statusText = MutableStateFlow("No images found")
+  private val _dropDownUiState = MutableStateFlow(emptyList<String>())
+  val dropDownUiState = _dropDownUiState.asStateFlow()
+  val shouldSeeIndex = MutableStateFlow(-1)
+  private var updateListJob: Job? = null
+  private val gradleTask = RoborazziGradleTask()
+
+  fun onRefreshClicked(project: Project) {
+    roborazziLog("onRefreshClicked")
+    refreshList(project)
+  }
+
+  fun onInit(project: Project) {
+    roborazziLog("onInit")
+    refreshList(project)
+  }
+
+  fun onSelectedFileChanged(project: Project) {
+    roborazziLog("onSelectedFileChanged")
+    coroutineScope.launch {
+      updateListJob?.cancel()
+      refreshListProcess(project)
+      selectListIndexByCaret(project)
+      fetchTasks(project)
+    }
+  }
+
+  fun onCaretPositionChanged(project: Project) {
+    roborazziLog("onCaretPositionChanged")
+    coroutineScope.launch {
+      updateListJob?.cancel()
+      refreshListProcess(project)
+      selectListIndexByCaret(project)
+      fetchTasks(project)
+    }
+  }
+
+  fun executeTaskByName(project: Project, taskName: String) {
+    roborazziLog("executeTaskByName")
+    gradleTask.executeTaskByName(project, taskName)
+  }
+
+  private fun selectListIndexByCaret(project: Project) {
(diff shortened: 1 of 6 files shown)


eyedol:
> Regarding the UI: I think it would be beneficial to have a reload button or a clear way to run tasks, rather than running them through selection.

Oh you want to discard the drop-down approach? I'm not sure how the reload button would work. Like you want me to put a reload in-place of the drop-down menu then when pressed, it shows the drop-down menu to select a roborazzi task to run? What will the reload button do? Is that to refresh the images found and also to refresh the task drop-down list? 

> Regarding the code: Currently, the tasks are not updated after creation. I suggest fetching tasks in onSelectedFileChanged and onCaretPositionChanged in the ViewModel, saving them in StateFlow, and collecting them in RoborazziPreviewPanel to apply the task list from the currently selected file.

Yes. I was going to move it to the viewmodel. I wasn't sure where exactly to be calling it to refresh. Thanks for suggesting where to call it.

takahirom:
Sorry the comment was confusing. 

I mean I wanted to add a refresh button along with selection menu. Like this:

```
[recrodRoborazziDebug] 🔄
```

You can select Gradle task with selection menu and you can run Gradle task by refresh button.

eyedol:
Oh I see and no worries. So you still want the drop-down menu but with icon support. Upon clicking on the icon, you execute the selected task. Right? 

If you don't mind sharing why you only want to run the task when hitting the refresh button from the selected task and not when selecting it, that would be great because I think with this approach we are introducing another level of expectation on the user to cause a selected task to run.

Thanks for clarifying my earlier response.

takahirom:
You can run the Gradle tasks when you select them, but I was thinking of following Android Studio's default configuration menu and run buttons behavior, which don't run tasks immediately upon selection. This approach might be more familiar to users.

takahirom:
However, I feel there's still a technical issue with obtaining the task that should be selected by default, and in that case, users would need to choose a task every time they start Android Studio. Therefore, executing upon selection might make sense.

eyedol:
I see what you mean. The AS example made it so clear to me. Thank you. I guess we could persist the selected task with `PropertiesComponent` then restore the selection upon creating the task drop-down menu.

eyedol:
Now we have this. Like you suggested right?

Screenshot 2024-07-12 at 8 38 42 PM

roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/RoborazziGradleTask.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.jetbrains.plugins.gradle.util.GradleUtil
+
+class RoborazziGradleTask {
+
+  fun fetchTasks(project: Project): List<String> {
+    val gradleModuleData = getGradleData(project) ?: return emptyList()
+
+    return ExternalSystemApiUtil.findAll(gradleModuleData, ProjectKeys.TASK)
+      .filter { it.data.name.contains("Roborazzi", true) && it.data.name.contains("DirRoborazzi", true).not()}
+      .map { it.data.name }.sorted()
```
takahirom, line 28: I would like to refresh only the current module, as it could take a lot of time to build all modules. How about using gradleModuleData.data.id?

```suggestion
      .map { gradleModuleData.data.id + ":" + it.data.name }.sorted()
```

roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/RoborazziPreviewTool.kt
```diff
@@ -86,222 +65,11 @@ class RoborazziPreviewToolWindowFactory : ToolWindowFactory {
...
-}
-
 class RoborazziPreviewPanel(project: Project) : JPanel(BorderLayout()) {
   private val listModel = DefaultListModel<Pair<String, Long>>()
-  private val statusLabel = JBLabel("No images found")
+  private val statusGradleTaskPanel = StatusToolbarPanel(project) { taskName ->
+    viewModel?.executeTaskByName(project, taskName)
+  }
   private val statusBar = JBBox.createHorizontalBox().apply {
     add(JLabel("Refresh: ").apply {
```
takahirom, line 74: I think we no longer need this JLabel("Refresh: "). Thanks to the new refresh button 🙏

roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/PreviewViewModel.kt
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  fun executeTaskByName(project: Project, taskName: String) {
+    roborazziLog("Executing task '$taskName'...")
+    _dropDownUiState.update { currentUiState ->
+      currentUiState.copy(flag = ActionToolbarUiState.Flag.LOADING)
+    }
+    gradleTask.executeTaskByName(project, taskName) {
+      _dropDownUiState.update { currentUiState ->
+        currentUiState.copy(flag = ActionToolbarUiState.Flag.IDLE)
```
takahirom, line 79: Could we add `refreshList(project)` here to update images after the Gradle task?

takahirom (approved):
I have some small points. But it looks great and very useful!

eyedol:
Let me know if you need any documentation add or update. I believe the screenshot page on the [IntelliJ](https://plugins.jetbrains.com/plugin/24561-roborazzi) plugin page needs updating.

takahirom:
Looks great! Thanks for your awesome contribution! I've made a few minor adjustments. Could you please take a look at the new PR?

PR #430, opened 2024-07-11
Dump both content description as well as text for the case when two elements are merged

lukas-mercari:
When dumping accessibility texts, I noticed that we either included content description or text.

However, it seems like when you merge e.g. an `Icon` with a content description and a `Text` field, both of them are included. So for those cases, only the content description would be included in our regression tests.

Note: I didn't include the same fix for the legacy view system, since I'm not sure if it also applies there.


takahirom (approved):
Looks great! Thanks for improving our Dump.AccessibilityExplanation!

PR #443, opened 2024-07-23
Tweak idea plugin UI

takahirom:
The Gradle tasks used to shrink like this:
image

We move the status label to the bottom
image

### DIFF
--- a/roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/TaskToolbarPanel.kt
+++ b/roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/TaskToolbarPanel.kt
@@ -15,34 +15,25 @@ import com.intellij.openapi.actionSystem.impl.ActionToolbarImpl
 import com.intellij.openapi.project.DumbAwareAction
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Key
-import com.intellij.ui.components.JBLabel
+import com.intellij.ui.components.JBBox
 import com.intellij.ui.components.JBPanel
 import com.intellij.util.ui.JBUI
 import java.awt.Dimension
 import java.awt.GridBagConstraints
 import java.awt.GridBagLayout
 import javax.swing.BorderFactory
-import javax.swing.Box
 import javax.swing.JComponent
 
-class StatusToolbarPanel(
+class TaskToolbarPanel(
   project: Project,
   onTaskExecuteButtonClicked: (taskName: String) -> Unit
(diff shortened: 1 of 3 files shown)


roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/TaskToolbarPanel.kt
```diff
@@ -52,31 +43,20 @@ class StatusToolbarPanel(
...
+    val actionToolbarConstraints = GridBagConstraints().apply {
+      insets = JBUI.insets(4)
+      gridx = 0
+      gridy = 0
+      anchor = GridBagConstraints.WEST
     }
-    add(_statusLabel, gbc)
-    val hRGlueConstraints = gbc.apply {
+    add(actionToolbar.component, actionToolbarConstraints)
+    add(JBBox.createHorizontalGlue(), GridBagConstraints().apply {
```
eyedol, line 54: I believe now that we have the status label moved to the bottom we could just use a `BorderLayout` so we don't have to set all these constraints and the horizontal glue.

Minor: Also I foresee the border still showing when the `actionToolbar` visibility is toggled to `false`. We could toggle the JBox instead that's if we wrap the `actionToolbar` into the JBBox's horizontal box. Something along the lines:

```kotlin
...
 private val jbBox = JBBox.createHorizontalBox()

  init {
    actionToolbar.setTargetComponent(this)
    add(jbBox.apply {
      isVisible = false
      setBorder(BorderFactory.createEmptyBorder(4, 4, 4, 4))
      add(actionToolbar.component, GridBagConstraints().apply {
        gridx = 0
        gridy = 0
        anchor = GridBagConstraints.WEST
        insets = JBUI.insets(4)
      })
    }, BorderLayout.WEST)
  }

  fun setActions(actions: List<ToolbarAction>) {
    if (actions.isEmpty()) {
      jbBox.isVisible = false
      return
    }
    jbBox.isVisible = true
    val actionList = listOf(*actions.toTypedArray())
    actionToolbar.setActions(actionList)
  }
...  
```

takahirom: Thanks for your advice. I released it a few days ago. I want to include these changes, but I'm a little busy. I'll do this the next time I modify the plugin, but if you could do it, it would be highly appreciated.

eyedol: Sure. I'll look into it.

PR #445, opened 2024-07-23
Support includePrivatePreviews of ComposablePreviewScanner

### LINKED ISSUE #441: [Feature Request] Private Preview Annotation Support
I would like to request support for private previews.
I suggest adding an option to use `includePrivatePreviews()` when working with ComposablePreviewScanner.

Reference Links:
- [ComposablePreviewScanner's Documentation](https://github.com/sergio-sastre/ComposablePreviewScanner/blob/4d899cfe6e5243bb03f8b1381719581423de6e40/README.md?plain=1#L86)
- [Roborazzi Code](https://github.com/takahirom/roborazzi/blob/b10e3e96f8f2fbdc47c3ce632e42192aa7561fed/roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt#L45)

### LINKED ISSUE #446: Compose Preview friction
https://takahirom.github.io/roborazzi/preview-support.html#generate-compose-preview-screenshot-tests

seems incomplete, running with 1.23.0 my gradle sync fails with 

```
Please set roborazzi.generateRobolectricPreviewTests.packages in the generatePreviewTests extension or set roborazzi.generateRobolectricPreviewTests.enable = false.See https://github.com/sergio-sastre/ComposablePreviewScanner?tab=readme-ov-file#how-to-use for more information.
```

Could this fail at runtime instead?  And should it be in the docs?

takahirom:
close #441
close #446


include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
```diff
@@ -48,6 +48,14 @@ open class GenerateComposePreviewRobolectricTestsExtension @Inject constructor(o
           "qualifiers" to "RobolectricDeviceQualifiers.Pixel4a",
         )
       )
+
+  /**
+   * The fully qualified class name of the custom test class that implements [com.github.takahirom.roborazzi.ComposePreviewTester].
+   * This is advanced usage. You can implement your own test class that implements [com.github.takahirom.roborazzi.ComposePreviewTester].
+   */
+  val testerQualifiedClassName: Property<String> = objects.property(String::class.java)
```
takahirom, line 56: I've simply changed the order because testerQualifiedClassName is an advanced option.

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -19,13 +19,24 @@ fun ComposablePreview<AndroidPreviewInfo>.captureRoboImage(
 
 @ExperimentalRoborazziApi
 interface ComposePreviewTester<T : Any> {
+  class Options(
```
takahirom, line 22: This would be the biggest change for users in this PR. This change is a breaking change for users using ComposePreviewTester. This is for reducing future breaking changes when we add options, as adding parameters to the interface would break users' implementations. Do you have any opinions about this?

takahirom:
@sergio-sastre Thanks as always. I've added support for includePrivatePreviews. I also did a little refactoring to avoid making breaking changes. I would appreciate your opinion. Do you have time to review this?

sergio-sastre:
> @sergio-sastre Thanks as always. I've added support for includePrivatePreviews. I also did a little refactoring to avoid making breaking changes. I would appreciate your opinion. Do you have time to review this?

Sure! I‘ll find some time today in the evening 😉

README.md
```diff
@@ -994,13 +994,13 @@ roborazzi {
...
 
-Roborazzi provides a helper function for ComposePreviewScanner.
+Roborazzi provides a helper function for ComposablePreviewScanner.
 You can add the following dependency to your project to use the helper function:
 
 `testImplementation("io.github.takahirom.roborazzi:roborazzi-compose-preview-scanner-support:[version]")`
 
 Then you can use the `ComposablePreview<AndroidPreviewInfo>.captureRoboImage()` function to capture the Composable Preview using the settings in Preview annotations.
-To obtain the `ComposablePreview` object, please refer to [ComposePreviewScanner](https://github.com/sergio-sastre/ComposablePreviewScanner).
+To obtain the `ComposablePreview` object, please refer to [ComposablePreviewScanner](https://github.com/sergio-sastre/ComposablePreviewScanner).
```
sergio-sastre, line 1003: I named it like that and I had my own doubts.
Google is also inconsistent on whether to call them "Compose Previews", like in "Compose Preview Screenshot Testing tool" or "Composable Previews" like [here](https://developer.android.com/develop/ui/compose/tooling/previews)

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
sergio-sastre: I think it does not require a breaking change.

You could simply extend the interface with
```kotlin
fun getPreviewOptions(): Options
```
and then use that new method in `ComposePreviewTester<AndroidPreviewInfo>.previews()`

Or I am missing something?

README.md
takahirom: Maybe I know it should not be `Compose Composable Preview`. 😊

> "A good name is long enough to fully communicate what the item is or does, without being so long that it becomes hard to read." https://google.github.io/eng-practices/review/reviewer/looking-for.html

I personally think we can use ComposePreview in an Android context, and also I think we can use ComposablePreview in a Jetpack Compose context. This is because Composable is an annotation name, so it's a minor distinction, while Compose is the framework name.

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -19,30 +21,59 @@ fun ComposablePreview<AndroidPreviewInfo>.captureRoboImage(
...
 @ExperimentalRoborazziApi
 interface ComposePreviewTester<T : Any> {
+  data class Options(
+    val testLifecycleOptions: TestLifecycleOptions = JUnit4TestLifecycleOptions(),
+    val scanOptions: ScanOptions = ScanOptions(emptyList()),
+  ) {
+    interface TestLifecycleOptions
+    class JUnit4TestLifecycleOptions(
+      // Used from generated tests
+      @Suppress("unused") val testRule: TestRule = object : TestWatcher() {},
```
takahirom, line 31: @yschimke 
Thanks as always. It's a little unrelated to this PR, but to address the feedback, I introduced a test rule option to support lifecycle hooks in Preview Support. Could you take a look?
https://github.com/takahirom/roborazzi/issues/446#issuecomment-2247203630

takahirom:
> You could simply extend the interface with
fun getPreviewOptions(): Options = defaultOptionsHere

Thank you for the great advice. I applied it, and it allowed us to pass the JUnit test rules from users.

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
yschimke: I'd like to see a sample of using it.

Does every test get it's own instance of TestRule?

They can be stateful https://github.com/junit-team/junit4/blob/ed47b7f487bafa48cff47f051af81a004cd36049/src/main/java/org/junit/rules/ErrorCollector.java

takahirom: Thanks for your great review! I changed testRule to testRuleFactory to create a test rule for each test and added integration tests. Could you check it out?
https://github.com/takahirom/roborazzi/pull/445/commits/9e20058c9daefbdc44b9752bb974a0cb61b4860e#diff-caa31d18b2a269ce09960b3943b693c5dcdd8c57ee34cd096b4a7ce27b6423adR8

yschimke: Yeah, I think that looks good

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -27,16 +27,31 @@ interface ComposePreviewTester<T : Any> {
   ) {
     interface TestLifecycleOptions
     data class JUnit4TestLifecycleOptions(
+      /**
+       * The TestRule factory to be used for the generated tests.
+       * You can use this to add custom behavior to the generated tests.
+       */
       // Used from generated tests
       @Suppress("unused") val testRuleFactory: () -> TestRule = { object : TestWatcher() {} },
```
yschimke, line 35: I do wonder whether there is some context that should be passed into the testRuleFactory? Or the methods on TestRule are enough?

takahirom: I thought the most likely scenario would be that we need to pass ComposeRule. I made adjustments to enable passing the ComposeRule.
https://github.com/takahirom/roborazzi/pull/445/commits/2f1255157b7b7c0180ab7116744db149f9acda06#diff-caa31d18b2a269ce09960b3943b693c5dcdd8c57ee34cd096b4a7ce27b6423adR13

yschimke: That looks like a great option. I've frequently needed to use the ComposeTestRule to interact with the UI.  nice.

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -16,33 +18,83 @@ fun ComposablePreview<AndroidPreviewInfo>.captureRoboImage(
...
+  fun options(): Options = defaultOptionsFromPlugin
+
   /**
    * Retrieves a list of composable previews from the specified packages.
    *
    * @param packages Vararg parameter representing the package names to scan for previews.
    * @return A list of ComposablePreview objects of type T.
    */
-  fun previews(vararg packages: String): List<ComposablePreview<T>>
+  fun previews(): List<ComposablePreview<T>>
```
takahirom, line 67: This is going to be a breaking change. I considered providing a default method, but I think it would make it difficult for users to know what to do.

takahirom: The release note will be like this.

## Breaking changes for users of the `ComposePreviewTester` interface

We've released [Compose Preview Support](https://takahirom.github.io/roborazzi/preview-support.html) in version 1.22.0. You can now modify the behavior of Compose Preview Support using ComposePreviewTester. Previously, the API of ComposePreviewTester was prone to breaking changes with each new option added. We've addressed this issue by introducing a new `options()` function. However, this introduces a breaking change for current users of ComposePreviewTester.

Old interface:
```kotlin
fun previews(vararg packages: String): List<ComposablePreview<T>>
```

New interface (Now you can access packages via `options().scanOptions.packages`):
```kotlin
fun previews(): List<ComposablePreview<T>>
```

## New feature: Support for includePrivatePreviews in Compose Preview Support

The `includePrivatePreviews` option allows you to include private previews in Compose Preview Support. You can set `includePrivatePreviews` in `roborazzi.generateComposePreviewRobolectricTests.includePrivatePreviews`.

## New feature: JUnit rule support in ComposePreviewTester

Previously, ComposePreviewTester lacked lifecycle hooks, making it challenging to handle certain scenarios. Now, you can pass your own Test rules, including your Compose Test Rule, and use them in tests. For a sample implementation, refer to [this integration test](https://github.com/takahirom/roborazzi/blob/2519f46ec0e5a8dadc5139ae833e2925c90c8b21/include-build/roborazzi-gradle-plugin/src/integrationTest/projects/sample-generate-preview-tests/src/test/java/com/github/takahirom/sample/CustomPreviewTester.kt).

takahirom:
Thank you for your reviews. I'll merge and release this.

PR #455, opened 2024-07-29
Don't crop dropdown task run button

### LINKED ISSUE #454: Fix issue where task run button is cropped in IDE plugin
image

eyedol:
Fixes #454 

- Remove anchoring of the JBBox harboring the toolbar to `BorderLayout.WEST` as it causing the user to resize the window to reveal it fully.

## Screenshot 
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/18d34773-0b79-4c30-834b-da03d84b6c3e" width="300" alt="Screen recording showing issue with toolbar dropdown hiding and requiring a window resize" /> | <video src="https://github.com/user-attachments/assets/b0732be7-be97-4ae7-a44e-b6aa89619603" width="300" alt="Screen recording showing the issue fixed where the toolbar dropdown doesn't require window resizing" />


takahirom:
Thanks! Looks better! Ideally, I think it would be good to see weight 1 behavior for task names like :feature:verifyAnd... ♻️ (the 'run task' button should be wrap_content and the task name should have weight 1). However, this seems difficult to implement.

PR #462, opened 2024-08-04
Compatibility check for Kotlin 2.0

takahirom:


sergio-sastre:
@takahirom 
ComposablePreviewScanner 0.7.0 now supports all previewInfo parameters of Compose Multiplatform Previews in common.

However, such Previews require Kotlin 2.0.0.
In order to update the example for Compose Multiplatform Previews, I have forked from this branch:
https://github.com/sergio-sastre/roborazzi/tree/droidcon/preview_tests/sample-generate-preview-tests-multiplatform

I'm wondering whether it would be possible to update only the :sample-generate-preview-tests-multiplatform module to kotlin 2.0.0 to update the sample also in Roborazzi's repo itself?

takahirom:
@sergio-sastre  Thank you for your feedback. I just don't want to inconvenience users, but I think we can update to Kotlin 2.0.0 because quite some time has passed since its release.

takahirom:
@sergio-sastre  I have one additional question that I’m still not clear about. Since Roborazzi is basically a library, I assume that updating the version on the application side will also update the Kotlin version and so on. Was there actually any issue—for example, Roborazzi’s bytecode still invoking APIs that have been removed or otherwise referencing deprecated/old APIs?

sergio-sastre:
> Ah, I think I understand now: we can’t add the new Compose Preview sample to the Roborazzi repo, at least for now.

Yes exactly. Otherwise, no issues 😊

PR #464, opened 2024-08-04
 Set a classDiscriminator to avoid polymorphic serializer error

### LINKED ISSUE #463: Roborazzi iOS fails to run due to the error: "Polymorphic serializer for class com.github.takahirom.roborazzi.CaptureResult.Recorded has property 'type' that conflicts with JSON class discriminator"
## What happened

I noticed that roborazzi-compose-ios has an error, although CI is passing. However, the CI is actually being skipped. The reason for this is that GitHub has switched macos-latest to arm64, and we are running `compareRoborazziIosX64`. We may need to add `compareRoborazziIosArm64` to CI and fix the issue.
I ran git bisect and identified the first commit causing the error as 217b53d8621d0bcc80123830d5db967ae1148879.

## How to reproduce

```
./gradlew compareRoborazziIosArm64 compareRoborazziIosX64
```

eyedol:
Fixes #463

### DIFF
--- a/.github/workflows/CompareScreenshot.yml
+++ b/.github/workflows/CompareScreenshot.yml
@@ -56,7 +56,7 @@ jobs:
       - name: compare screenshot test
         id: compare-screenshot-test
         run: |
-          ./gradlew compareRoborazziDebug compareRoborazziDesktop compareRoborazziJvm compareRoborazziIosX64  --stacktrace --info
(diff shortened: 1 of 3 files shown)


eyedol:
@takahirom It seems the Json configured in `RoborazziIos#writeJson` has no `classDiscriminator` defined hence the PolymorphicSerializer serializer failure from ksx. I reused the same json instance from `CaptureResults#Companion#json` configuration which has a `classDiscriminator` defined and it seems to have resolved it.

takahirom:
Thank you for your prompt contribution!

Issue #460, opened 2024-07-31
Get diff value in test summary


vladcudoidem:
Are there plans to include the diff value in the test summary JSON? Currently, there seems to be no way of telling what the difference percentage is. Or am I missing something?

takahirom:
Thank you for your suggestion. I think it would be a valuable addition. However, may I ask about the specific use case for including the difference percentage, as opposed to setting a tolerance for the diff?

vladcudoidem:
Sure. I do not have a specific use case in mind, but I am looking into large scale automated screenshot testing and now especially into what one can do with long term results (e. g. statistical analysis). Currently, the data that one gets is binary (passed or failed) and I think it would be nice to be able to get more information, like diff percentages. (I have not looked into the inner workings of Roborazzi, but I suppose that you calculate the total diff for any given screenshot anyway.) Also, it would be nice to be able to display such information in PRs or in management tools.

Actually, here's one use case: say one has to set a high threshold for a particular test (maybe because of flakiness or platform-dependent rendering), for example 6%. The test passes for both 0.1% and 5.5% difference. While the 0.1% is irrelevant, 5.5% might be a critical UI regression that is not caught because the arbitrary value of 6% was chosen a little too high and misses the 4% to 5.9% range that might be critical. There might also be a categorical difference between a fail with a 6.5% diff and one with a 80% diff. To put it differently, binary information sometimes does not deliver enough information. I think it is obvious that these are edge cases, but I think that they would unlock a few interesting evaluation and automation possibilities. Also, I see no obvious reason why outputting more information would be bad in any way. I hope I'm not missing something.

Thanks for the consideration!

(Btw, just curious, do you think that implementing this would require a large amount of effort? In other words, are a lot of changes necessary? If not, I could maybe attempt to implement this myself.)

takahirom:
The amount of code we might need to change is relatively small, but the code here is somewhat complex due to various use cases.
https://github.com/takahirom/roborazzi/blob/main/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.kt#L88C11-L88C27

You can obtain a ComparisonResult, which includes pixelDifferences and pixelCount. 
https://github.com/dropbox/differ/blob/39d931731482587c1ba34a58412aa8af65add230/differ/src/commonMain/kotlin/com/dropbox/differ/ImageComparator.kt#L7

[...]

PR #480, opened 2024-09-11
Implement diff. percentage

vladcudoidem:
**This PR is my proposed solution for #460.**

Please take a look at the way I implemented the property `diffPercentage` in the `CaptureResult.Changed` data class. Another possibility would be to declare `diffPercentage` directly as a property in the `CaptureResult` interface, and implement them as null-returning-getters in the other data classes, as you have done for other properties. (This would also remove the need for my if-statement in `CaptureResultTest.kt`.)

Also, as far as I can tell, you plan to use the Dropbox Differ in the future for iOS as well (see [this](https://github.com/takahirom/roborazzi/blob/1924d408d2e99e1d9e65a557d7f3711fa9368367/roborazzi-compose-ios/src/iosMain/kotlin/io/github/takahirom/roborazzi/RoborazziIos.kt#L366) comment). That's why I have implemented the `diffPercentage` to always equal `Float.NaN` (undefined) for iOS (the reason being that your diffing algorithm for iOS would require a lot of changes to be able to deliver the needed data for calculating a diff. percentage).

I hope you like this contribution :)
Don't hesitate to tell me if I need to change something.

### DIFF
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.kt
@@ -2,6 +2,7 @@ package com.github.takahirom.roborazzi
 
 import com.dropbox.differ.ImageComparator
 import java.io.File
+import kotlin.properties.Delegates
 
 fun interface EmptyCanvasFactory {
   operator fun invoke(
@@ -84,6 +85,10 @@ fun processOutputImageAndReport(
         bufferedImageType = recordOptions.pixelBitConfig.toBufferedImageType()
       )
     }
+
+    // Only used by CaptureResult.Changed
+    var diffPercentage by Delegates.notNull<Float>()
+
     val changed = if (height == goldenRoboCanvas.height && width == goldenRoboCanvas.width) {
       val comparisonResult: ImageComparator.ComparisonResult =
         newRoboCanvas.differ(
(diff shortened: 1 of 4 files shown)


vladcudoidem:
Also, I would love to include this contribution as part of my bachelor's thesis about screenshot testing.

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.kt
```diff
@@ -84,17 +85,23 @@ fun processOutputImageAndReport(
         bufferedImageType = recordOptions.pixelBitConfig.toBufferedImageType()
       )
     }
+
+    // Only used by CaptureResult.Changed
+    var diffPercentage by Delegates.notNull<Float>()
```
takahirom, line 90: I just want to know this: the reason you use `Delegates.notNull<Float>()` instead of `var diffPercentage = NaN` is that you want to trigger an error when the value is not set, right?

vladcudoidem: Yes, I think it is a little more robust to force the developer to set the `diffPercentage` in each case separately than to set a default and only change it in the cases where it should be different. But it is mostly a matter of taste, in my opinion. I can change it if you wish.

vladcudoidem:
Also, I did not log anything, because I wanted to know from you if you want the `diffPercentage` logged anywhere or not. I can add it if you want.

takahirom (approved):
I'm sorry for being late. Thank you for your contribution.

vladcudoidem:
No worries. Thank you as well!

PR #494, opened 2024-09-26
fix duplicate screenshot

yuchan2215:
For example, let's say there are the following functions in the file: `ExampleChip` and `ExampleChipPreview`. In the current implementation, `ExampleChip` is also included in `ExampleChipPreview`, resulting in the same preview being displayed twice. To avoid this, I added a process to delete duplicate files.

https://github.com/takahirom/roborazzi/blob/fe9d08a77197055138c58189ccd21764363c9e4a/roborazzi-idea-plugin/src/main/kotlin/com/github/takahirom/roborazzi/idea/preview/PreviewViewModel.kt#L263-L280

I am not very good at English, so I am using ChatGPT.


takahirom (approved):
Looks good. Thank you for your contribution!

PR #496, opened 2024-10-02
Initial integration of differ to support iOS

### LINKED ISSUE #383: Migrate iOS implementation to use RoboCanvas Interface
Implement iOS implementation of `RoboCanvas`. In other to accomplish this, the issue with differ library needs to be resolved. See: https://github.com/dropbox/differ/pull/16#issuecomment-2119081492

eyedol:
Fixes #383


takahirom (approved):
Looks great! Thank you for coordinating with other libraries and advancing our library foundation.
@eyedol, do you have anything to add to make this PR ready?

eyedol:
> Looks great! Thank you for coordinating with other libraries and advancing our library foundation.
> @eyedol, do you have anything to add to make this PR ready?

I was going to look into how to use `RoboCanvas` to make an iOS implementation similar to how it's done with `AwtRoboCanvas` however, I couldn't figure out where to begin. Basically I wanted to work on the comments in `RoborazziIos.kt`

> This is a temporary implementation for iOS.
> We need to transition these implementations to Kotlin Multiplatform and RoboCanvas....

If you could point me in the right direction what you wanted to achieve with the comments in there, that would be great. Also if you think it will be a huge change in the end, we can tackle it with another PR. Let me know. Thanks.

takahirom:
I think gradually moving the code to `commonMain` without using it for iOS will allow us to make progress without bugs, and then we can finally use it for iOS.

takahirom:

PR #529, opened 2024-11-01
Add the ability to change File writers and readers and add support for lossless WebP

### LINKED ISSUE #522: Different captured image format
Hello there
Does Roborazzi support capturing\comparing in other formats rather than PNG (i have a particular interest in webp since it is lighter)?
Thanks

takahirom:
close #522

### DIFF
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/ImageIoFormat.commonJvm.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/ImageIoFormat.commonJvm.kt
@@ -100,7 +100,7 @@ fun ImageWriter.writeMetadata(
 /**
  * Add testImplementation("io.github.darkxanter:webp-imageio:0.3.3") to use this
  */
-fun losslessWebPWriter(): AwtImageWriter =
+private fun losslessWebPWriter(): AwtImageWriter =
   AwtImageWriter { file, context, bufferedImage ->
     val writer: ImageWriter =
       ImageIO.getImageWritersByMIMEType("image/webp").next()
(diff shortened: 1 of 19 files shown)


include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/ImageIoFormat.commonJvm.kt
```diff
@@ -0,0 +1,123 @@
...
+    text.appendChild(textEntry)
+    root.appendChild(text)
+  }
+
+  meta.mergeTree(IIOMetadataFormatImpl.standardMetadataFormatName, root)
+  return meta
+}
+
+/**
+ * Add testImplementation("io.github.darkxanter:webp-imageio:0.3.0") to use this
```
ArcaNO93, line 101: every1 who wants to use webp format should add this?
is so, why not adding it as transitive dependency?

takahirom: Thanks, that makes sense. However, since we need to download additional code when using this, there are trade-offs to consider. Additionally, the roborazzi-core module is already being utilized by the Roborazzi Gradle Plugin. While we may need to add the roborazzi-webp module, it could complicate maintenance. This is why I'm using a compileOnly dependency. It would be beneficial to add a comment explaining the compileOnly dependency.
image

ArcaNO93: makes sense, yeah
should be enough for now to have it in readme

ArcaNO93 (approved):

PR #530, opened 2024-11-01
Use RobolectricDeviceQualifierBuilder from ComposablePreviewScanner to support preview info "device"

sergio-sastre:


roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RobolectricPreviewInfosApplier.kt
```diff
@@ -4,12 +4,17 @@ import android.content.res.Configuration
...
 import org.robolectric.RuntimeEnvironment.setQualifiers
 import sergio.sastre.composable.preview.scanner.android.AndroidPreviewInfo
+import sergio.sastre.composable.preview.scanner.android.device.domain.RobolectricDeviceQualifierBuilder
 import sergio.sastre.composable.preview.scanner.core.preview.ComposablePreview
 
 @ExperimentalRoborazziApi
 fun ComposablePreview<AndroidPreviewInfo>.applyToRobolectricConfiguration() {
   val preview = this
 
+  RobolectricDeviceQualifierBuilder.build(preview.previewInfo.device)?.run {
```
sergio-sastre, line 14: @takahirom 
The only trouble that I see is how to handle the "default device" from the Roborazzi plugin option.

AndroidPreviewInfo provides "" as default device, which is the same as the `@Preview(device = Devices.DEFAULT)`, and this would override the one in the Roborazzi plugin option... how would you like to handle that?

takahirom: Is this the 'default device' you mentioned? 

```kotlin
  /**
   * [robolectricConfig] will be passed to the Robolectric's @Config annotation in the generated test class.
   * See https://robolectric.org/configuring/ for more information.
   */
  val robolectricConfig: MapProperty<String, String> = objects.mapProperty(String::class.java, String::class.java)
    .convention(
      mapOf(
        "sdk" to "[33]",
        "qualifiers" to "RobolectricDeviceQualifiers.Pixel4a",
      )
    )
```

For the basic strategy, settings with a smaller scope should take priority. Therefore, I think we should check if it's blank and leave it as is if there's no setting.

```kotlin
if(!preview.previewInfo.device.isBlank()) {
  RobolectricDeviceQualifierBuilder.build(preview.previewInfo.device)?.run {
     setQualifiers(this)
  }
}
```

sergio-sastre: Yes, that is what I mean.

I actually believe that is the best solution and makes sense.

I’ll change it accordingly

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RobolectricPreviewInfosApplier.kt
```diff
@@ -4,12 +4,22 @@ import android.content.res.Configuration
...
+import sergio.sastre.composable.preview.scanner.android.device.domain.RobolectricDeviceQualifierBuilder
 import sergio.sastre.composable.preview.scanner.core.preview.ComposablePreview
 
 @ExperimentalRoborazziApi
 fun ComposablePreview<AndroidPreviewInfo>.applyToRobolectricConfiguration() {
   val preview = this
 
+  fun setDevice(device: String){
+    if (device.isNotBlank()) {
+      RobolectricDeviceQualifierBuilder.build(device)?.run {
```
takahirom, line 16: Users may encounter compiler errors when updating Roborazzi. Since version management in Roborazzi Gradle Plugin can be challenging, how about adding a comment like this:

```suggestion
      // Requires `io.github.sergio-sastre.ComposablePreviewScanner:android:0.4.0` or later
      RobolectricDeviceQualifierBuilder.build(device)?.run {      ```

Issue #444, opened 2024-07-23
About deprecation of context receiver


takahirom:
We are using [context receiver](https://github.com/takahirom/roborazzi/blob/main/roborazzi-compose-desktop/src/commonMain/kotlin/io/github/takahirom/roborazzi/RoborazziDesktop.kt#L15) for desktop support.
I was thinking that we could migrate to context parameters. However, this might be challenging as it requires the latest Kotlin version, and we would need to upgrade all library versions, including Compose. While we could potentially make this change, library users might not be able to update. Therefore, I believe it's better to migrate to regular parameters for now.
https://youtrack.jetbrains.com/issue/KT-8087/Make-it-possible-to-suppress-warnings-globally-in-compiler-via-command-line-option#focus=Comments-27-10137847.0-0

PR #533, opened 2024-11-03
Try to remove context receiver of desktop

takahirom:
close https://github.com/takahirom/roborazzi/issues/444

### DIFF
--- a/README.md
+++ b/README.md
@@ -1204,13 +1204,6 @@ kotlin {
       }
     }
     ...
-
-// Roborazzi Desktop support uses Context Receivers
-    tasks.withType<KotlinCompile>().configureEach {
-      kotlinOptions {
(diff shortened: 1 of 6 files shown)


takahirom:
The reason we needed `context(DesktopComposeUiTest)` was that `compose/ui/ui-test/src/skikoMain/kotlin/androidx/compose/ui/test/SkikoImageHelpers.kt` didn't exist before, and we had to use DesktopComposeUiTest. However, now that we can use SkikoImageHelpers, we can remove the `context(DesktopComposeUiTest)`.

takahirom:
This feature was introduced in the Compose Multiplatform core repository (commit 1664fba). This indicates that the feature has been supported since Compose Multiplatform version 1.5.12.

PR #553, opened 2024-11-15
Add compose version of captureScreenRoboImage and use it in Roborazzi Compose Preview Support

takahirom:


takahirom:
I think this might be a wrong approach 😅

Issue #554, opened 2024-11-15
Dialog in Compose Preview causes `IllegalStateException` in generated tests


ashughes:
## Repro

Add the following to `Previews.kt` in `sample-generate-preview-tests`:

```kotlin
@Preview
@Composable
fun PreviewDialog() {
  MaterialTheme {
    AlertDialog(
      onDismissRequest = {},
      confirmButton = @Composable { Text("Confirm") },
      text = @Composable { Text("Generate Preview Test Sample!") }
    )
  }
}
```

Run `./gradlew :sample-generate-preview-tests:recordRoborazziDebug` and observe the following output:

```
RoborazziPreviewParameterizedTests > test[com.github.takahirom.preview.tests.PreviewsKt_PreviewDialog] FAILED
    java.lang.IllegalStateException at RoborazziPreviewParameterizedTests.kt:38

6 tests completed, 1 failed

> Task :sample-generate-preview-tests:testDebugUnitTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sample-generate-preview-tests:testDebugUnitTest'.
```

<details>
<summary>Stacktrace of failed test</summary>

```
java.lang.IllegalStateException: Unable to find the image of the target root component. Does the rendering element exist?
	at com.github.takahirom.roborazzi.RoborazziKt.capture(Roborazzi.kt:618)
	at com.github.takahirom.roborazzi.ImageCaptureViewAction.perform(Roborazzi.kt:592)
	at androidx.test.espresso.ViewInteraction$SingleExecutionViewAction.perform(ViewInteraction.java:2)
	at androidx.test.espresso.ViewInteraction.doPerform(ViewInteraction.java:25)
	at androidx.test.espresso.ViewInteraction.-$$Nest$mdoPerform(ViewInteraction.java)
	at androidx.test.espresso.ViewInteraction$1.call(ViewInteraction.java:7)
	at androidx.test.espresso.ViewInteraction$1.call(ViewInteraction.java:1)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at android.os.Handler.$$robo$$android_os_Handler$handleCallback(Handler.java:942)
	at android.os.Handler.handleCallback(Handler.java)
	at android.os.Handler.$$robo$$android_os_Handler$dispatchMessage(Handler.java:99)
	at android.os.Handler.dispatchMessage(Handler.java)
[...]

PR #555, opened 2024-11-15
Fix alert dialog support

takahirom:
close https://github.com/takahirom/roborazzi/issues/554


roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -169,6 +173,43 @@ fun captureScreenRoboImage(
...
+        "We can add a flag to disable this behavior so please let us know if you need it."
+    )
+    captureScreenRoboImage(file, roborazziOptions)
+  } else {
+    captureSingleComponent()
+  }
+}
+
+@InternalRoborazziApi
+fun fetchRobolectricWindowRoots(): List<Root> {
```
takahirom, line 195: This is the method we use for `captureRoboImage()` to fetch windows.

roborazzi/src/main/java/com/github/takahirom/roborazzi/Roborazzi.kt
```diff
@@ -169,6 +173,43 @@ fun captureScreenRoboImage(
   }
 }
 
+@InternalRoborazziApi
+fun captureScreenIfMultipleWindows(
```
takahirom, line 177: @sergio-sastre  
Thanks as always. I'm thinking about adding support for the `AlertDialog()` Composable.  
Currently, it throws an `IllegalStateException`, and I would like to fix that behavior.

The basic strategy here is to use `captureScreenRoboImage()` (which merges all windows into one screenshot) if there are multiple windows.
Do you have any thought on this?
image

sergio-sastre: @takahirom 
Oh sorry, I‘ve just seen this and also saw that it was merged.
I think it is fine as first solution.

Getting it right with the windows is a tricky one. 

Have you tried by converting the View to bitmap and capturing a bitmap from it? That would likely avoid the IllegalStateException. However, I am not sure whether we could use PixelCopy to draw the bitmap with elevation.

what I mean is to use sth like this `composeView.drawToBitmap.captureRoboImage()`

maybe worth giving it a try?

PR #562, opened 2024-11-21
Update accessibility sample to show presets

yschimke:


takahirom (approved):

Issue #569, opened 2024-11-22
Accessibility Checks warn on API < 34


yschimke:
The checks require API 34, but currently just silently skip if before that.

They should warn noisily when not running.

yschimke:
already there, but adding for native

PR #572, opened 2024-11-23
Log warnings when Accessibility checks not configured correctly

yschimke:
Add tests showing we don't fail when configured incorrectly, and log if not in NATIVE mode.

relates to https://github.com/takahirom/roborazzi/issues/569


takahirom (approved):

PR #573, opened 2024-11-23
Apply suppressions to AccessibilityViewCheckResult instead of filtering

### LINKED ISSUE #570: Log Accessibility checks suppressions
Log suppressions also.  Suppressions might be something to fix and should be visible in log.

Current logic filters them out as if they never happened.

yschimke:
This change introduces a new `applySuppressions` function to process suppressions for accessibility check results. It filters results based on suppression rules and adds a new `AccessibilityCheckResultType.SUPPRESSED` type for suppressed results. Additionally, it adds logging for suppressed results.

Existing logic filters instead of suppressing, making these invisible.

fixes #570

### DIFF
--- a/roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
+++ b/roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
@@ -175,8 +175,20 @@ data class RoborazziATFAccessibilityChecker(
 
     val results = viewChecker.runChecksOnView(ImmutableSet.copyOf(checks), view, parameters)
 
-    return results.filter {
-      !suppressions.matches(it)
+    return results.applySuppressions(suppressions)
+  }
+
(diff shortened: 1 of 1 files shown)


roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
```diff
@@ -180,8 +180,24 @@ data class RoborazziATFAccessibilityChecker(
...
+  private fun List<AccessibilityViewCheckResult>.applySuppressions(
+    suppressions: Matcher<in AccessibilityViewCheckResult>
+  ): List<AccessibilityViewCheckResult> {
+    val ranTypes = listOf(
+      AccessibilityCheckResultType.ERROR,
+      AccessibilityCheckResultType.WARNING,
+      AccessibilityCheckResultType.INFO
+    )
+
+    return map { checkResult ->
```
takahirom, line 195: I'm a little confused, so I added the parameter name. 👀

takahirom (approved):
Looks great! Thanks.

PR #574, opened 2024-11-23
Fix: Restore Robolectric fingerprint after accessibility checks

yschimke:
Restores the original Robolectric fingerprint after performing accessibility checks. This ensures that the fingerprint is not permanently modified, which can cause issues in other parts of the test suite.


takahirom (approved):
Test independence is important. Thanks!

yschimke:
@takahirom this wouldn't have survived across tests, but if you tried to scroll after running checks, you might hit some threading issues that compose works around.

PR #557, opened 2024-11-16
Add basic ATF A11y checks via AccessibilityCheckAfterTestStrategy and checkRoboAccessibility

yschimke:
Run ATF after test completes to check for A11y issues.

Future possible changes

 - Collect results and annotate screenshots with failures
 - demonstrate in CI showing results
 
 ```
  @get:Rule
  val roborazziRule = RoborazziRule(
    composeRule = composeTestRule,
    captureRoot = composeTestRule.onRoot(),
    options = Options(
      roborazziAccessibilityOptions = RoborazziATFAccessibilityCheckOptions(
        RoborazziATFAccessibilityChecker(
          preset = AccessibilityCheckPreset.LATEST,
          suppressions = matchesElements(withTestTag("suppress")),
        ),
        failureLevel = RoborazziATFAccessibilityChecker.CheckLevel.Warning,
      ),
      accessibilityCheckStrategy = AccessibilityCheckAfterTestStrategy()
    )
  )
 ```
 
 or
 
 ```
    composeTestRule.onNodeWithTag("contactcard").checkRoboAccessibility(
      roborazziATFAccessibilityCheckOptions = RoborazziATFAccessibilityCheckOptions(
        failureLevel = RoborazziATFAccessibilityChecker.CheckLevel.Warning
      )
    )
 ```

PR #575, opened 2024-11-23
Run accessibility checks after each screenshot

yschimke:
This commit introduces the ability to run accessibility checks after each screenshot is taken by Roborazzi. This is achieved by adding a new accessibility check strategy, `AccessibilityCheckEachScreenshotStrategy`, which is used by the `RoborazziRule`.

Additionally, the `captureAndroidView` and `captureComposeNode` functions have been updated to accept an `onEach` callback, which is invoked after each screenshot is taken. This callback can be used to perform any necessary actions, such as running accessibility checks.

Finally the checks are changed to only run when roborazzi is enabled and will be skipped completely if roborazzi is not running.

### DIFF
--- a/roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
+++ b/roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
@@ -10,6 +10,8 @@ import androidx.compose.ui.test.SemanticsNodeInteraction
 import androidx.test.espresso.UiController
 import androidx.test.espresso.ViewAction
 import androidx.test.espresso.ViewInteraction
+import com.github.takahirom.roborazzi.RoborazziRule.AccessibilityCheckStrategy
+import com.github.takahirom.roborazzi.RoborazziRule.AccessibilityCheckStrategy.CheckPoint
 import com.github.takahirom.roborazzi.RoborazziRule.CaptureRoot
 import com.google.android.apps.common.testing.accessibility.framework.AccessibilityCheckPreset
 import com.google.android.apps.common.testing.accessibility.framework.AccessibilityCheckResult.AccessibilityCheckResultType
@@ -228,14 +230,14 @@ data class RoborazziATFAccessibilityChecker(
   companion object
 }
 
-@ExperimentalRoborazziApi
-data class AccessibilityCheckAfterTestStrategy(
-  val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
-) : RoborazziRule.AccessibilityCheckStrategy {
+abstract class BaseAccessibilityCheckStrategy: AccessibilityCheckStrategy {
+  abstract val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions
+
+  private val accessibilityOptions by lazy { accessibilityOptionsFactory() }
+
   override fun runAccessibilityChecks(
     captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions
   ) {
-    val accessibilityOptions = accessibilityOptionsFactory()
     accessibilityOptions
       .checker
       .runAccessibilityChecks(
@@ -252,5 +254,23 @@ data class AccessibilityCheckAfterTestStrategy(
         roborazziOptions = roborazziOptions,
         failureLevel = accessibilityOptions.failureLevel,
       )
+  }
+}
+
+@ExperimentalRoborazziApi
+data class AccessibilityCheckEachScreenshotStrategy(
+  override val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
+) : BaseAccessibilityCheckStrategy() {
+  override fun shouldRunAt(checkPoint: CheckPoint): Boolean {
+    return checkPoint == CheckPoint.AfterScreenshot || checkPoint == CheckPoint.AfterTest
+  }
+}
+
+@ExperimentalRoborazziApi
+data class AccessibilityCheckAfterTestStrategy(
+  override val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
+) : BaseAccessibilityCheckStrategy() {
+  override fun shouldRunAt(checkPoint: CheckPoint): Boolean {
+    return checkPoint == CheckPoint.AfterTest
   }
 }
--- a/roborazzi-junit-rule/src/main/java/com/github/takahirom/roborazzi/RoborazziRule.kt
+++ b/roborazzi-junit-rule/src/main/java/com/github/takahirom/roborazzi/RoborazziRule.kt
@@ -3,6 +3,7 @@ package com.github.takahirom.roborazzi
 import androidx.compose.ui.test.SemanticsNodeInteraction
 import androidx.compose.ui.test.junit4.ComposeTestRule
 import androidx.test.espresso.ViewInteraction
+import com.github.takahirom.roborazzi.RoborazziRule.AccessibilityCheckStrategy.CheckPoint
 import org.junit.rules.TestWatcher
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
@@ -67,6 +68,13 @@ class RoborazziRule private constructor(
       captureRoot: CaptureRoot,
       roborazziOptions: RoborazziOptions,
     )
+
+    enum class CheckPoint {
+      AfterTest, AfterScreenshot
+    }
+
+    fun shouldRunAt(checkPoint: CheckPoint): Boolean
+
     // Use `roborazzi-accessibility-check`'s [com.github.takahirom.roborazzi.AccessibilityCheckAfterTestStrategy]
     data object None : AccessibilityCheckStrategy {
       override fun runAccessibilityChecks(
@@ -75,6 +83,8 @@ class RoborazziRule private constructor(
       ) {
         // Do nothing
       }
+
+      override fun shouldRunAt(checkPoint: CheckPoint): Boolean = false
     }
   }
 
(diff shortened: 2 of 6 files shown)


takahirom:
Thanks.  
This change might conflict with this PR. I'm still not sure if separating the stable constructor is a good approach. Do you have any thoughts on this? I would prefer to merge this first so we can determine if the change is treated as Experimental.  
https://github.com/takahirom/roborazzi/pull/566/files#diff-0bb7b68fe285cb6875132e144fbb4d226a3f35ff5525cd14d6c627f1389b5999R62-R63

yschimke:
Which one do you want to merge first?

Your call. I'll rebase later today and as necessary.

yschimke:
It would be good to avoid duplicate warnings, so in a follow up I might add some way to log only once for duplicate warning.

roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
```diff
@@ -247,14 +249,14 @@ data class RoborazziATFAccessibilityChecker(
   companion object
 }
 
-@ExperimentalRoborazziApi
-data class AccessibilityCheckAfterTestStrategy(
-  val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
-) : RoborazziRule.AccessibilityCheckStrategy {
+abstract class BaseAccessibilityCheckStrategy: AccessibilityCheckStrategy {
```
takahirom, line 252: Just a small point

```suggestion
@ExperimentalRoborazziApi
abstract class BaseAccessibilityCheckStrategy: AccessibilityCheckStrategy {
```

roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
```diff
@@ -273,3 +275,21 @@ data class AccessibilityCheckAfterTestStrategy(
       )
   }
 }
+
+@ExperimentalRoborazziApi
+data class AccessibilityCheckEachScreenshotStrategy(
+  override val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
+) : BaseAccessibilityCheckStrategy() {
+  override fun shouldRunAt(checkPoint: CheckPoint): Boolean {
```
takahirom, line 283: You might already be thinking this, but how about using interface functions directly?  

```kotlin
// AccessibilityCheckStrategy.kt
interface AccessibilityCheckStrategy {
    fun afterScreenshot(captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions) {}
    fun afterTest(captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions) {}
}

// BaseAccessibilityCheckStrategy.kt
abstract class BaseAccessibilityCheckStrategy : AccessibilityCheckStrategy {
    abstract val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions

    private val accessibilityOptions by lazy { accessibilityOptionsFactory() }

    protected fun runChecks(captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions) {
        accessibilityOptions
            .checker
            .runAccessibilityChecks(
                captureRoot = captureRoot,
                roborazziOptions = roborazziOptions,
                options = accessibilityOptions
            )
    }
}

// AccessibilityCheckEachScreenshotStrategy.kt
data class AccessibilityCheckEachScreenshotStrategy(
    override val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
) : BaseAccessibilityCheckStrategy() {
    override fun afterScreenshot(captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions) {
        runChecks(captureRoot, roborazziOptions)
    }

    override fun afterTest(captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions) {
        runChecks(captureRoot, roborazziOptions)
    }
}

// AccessibilityCheckAfterTestStrategy.kt
data class AccessibilityCheckAfterTestStrategy(
    override val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
) : BaseAccessibilityCheckStrategy() {
    override fun afterTest(captureRoot: CaptureRoot, roborazziOptions: RoborazziOptions) {
        runChecks(captureRoot, roborazziOptions)
    }
}

// RoborazziRule.kt modification
class RoborazziRule private constructor(
    // ... omitted ...
) {
    // ... omitted ...

    // Directly call the event-specific method when needed
    accessibilityChecks.afterScreenshot(captureRoot, options.roborazziOptions)

    // ... omitted ...

    // Also call the method at the end of the test
    accessibilityChecks.afterTest(captureRoot, options.roborazziOptions)
}
```

takahirom (commented):
Thank you for your pull request! Most parts look good.

roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
yschimke: Good idea.

roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
```diff
@@ -248,13 +249,14 @@ data class RoborazziATFAccessibilityChecker(
...
-data class AccessibilityCheckAfterTestStrategy(
-  val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions = { provideATFAccessibilityOptionsOrCreateDefault() },
-) : RoborazziRule.AccessibilityCheckStrategy {
-  override fun runAccessibilityChecks(
+abstract class BaseAccessibilityCheckStrategy : AccessibilityCheckStrategy {
+  abstract val accessibilityOptionsFactory: () -> RoborazziATFAccessibilityCheckOptions
+
+  private val accessibilityOptions by lazy { accessibilityOptionsFactory() }
+
+  fun runAccessibilityChecks(
```
takahirom, line 257:
```suggestion
  @InternalRoborazziApi
  fun runAccessibilityChecks(
```

roborazzi-accessibility-check/src/main/java/com/github/takahirom/roborazzi/RoborazziATFAccessibilityChecker.kt
```diff
@@ -273,3 +275,26 @@ data class AccessibilityCheckAfterTestStrategy(
       )
   }
 }
+
+@ExperimentalRoborazziApi
+data class AccessibilityCheckEachScreenshotStrategy(
```
takahirom, line 280: Thanks. I have a question about the current implementation. 
I haven't read the code thoroughly. The AccessibilityCheckEachScreenshotStrategy is only for RoborazziRule.Options.captureRule and not for captureRoboImage(), right? 
I believe that if we provide the RoborazziContext, we can check every captureRoboImage. Do you have any thoughts on this?

yschimke: I'm happy to take your advice, I'm not as familiar with all the different APIs.

yschimke: Really the intention is to run when roborazzi produces distinct screenshots.

From
- once per test
- per screenshot
- none (roborazzi or a11y checks not enabled)

What gets us closer to that?

takahirom: I think we can add `ruleAccessibilityCheckStrategy` and use it in `captureRoboImage()` like `provideRoborazziContext().accessibilityCheckStrategy.afterScreenshot()` to check the accessibility of each screenshot now that you added `onEach{}`.

https://github.com/takahirom/roborazzi/pull/557/commits/77fa9dc1dc9e68a797cc942405a9134dd52a27df#diff-84240ff001c21726c87098748f05f02f55cb70b8425874729662bc07ad8786ecR66

yschimke: Should I move that type from rule to core module?

takahirom: I've checked, and we can't use `CaptureRoot` because it is in a rule module. 👀

It's a little confusing, so I've renamed it to `RuleCaptureRoot`.

I think we can use `RoboComponent`, but it currently doesn't have fields for `View` or `SemanticsNode`.

However, I realize that this lazy block already retains the instance, and I'm using `RoboComponent.Screen` extensively. I believe we can add the fields to `RoboComponent.View` and `RoboComponent.Compose`. The `RoboComponent` should be collected by the garbage collector, so it should be fine.

So I think we can do it like this. But there might still be issues we have to address.

* Add `View` and `SemanticsNode` fields to `RoboComponent`

* Add `runAccessibilityChecks()` function to use `RoboComponent` (The `runAccessibilityChecks` might be needed for an after-test rule.)

* Change `afterScreenshot`'s parameter from `CaptureRoot` to `RuleCaptureRoot`

yschimke: This feels like a big change I won't get to until next weekend.

Options

- I'll take a look and make changes then
- feel free to make the changes yourself before then
- I can remove that screenshot mode, and just land the other changes for now

No rush, just letting you know where I am

yschimke: I started to make the change and noticed some are already there. I'm unclear if you made all the changes already?

But for RoboComponent, what to do about Screen? Since it has multiple roots?

takahirom: I'm not sure, but can we conduct checks for each root? I don't have a good idea for that, but it might work.

yschimke: Or just the top window?

takahirom: Sounds good. We could output logs like:
"Roborazzi: It seems we have multiple windows, so we'll check accessibility for the top window only."

Issue #559, opened 2024-11-17
Support @Preview params "widthDp" and "heightDp"


sergio-sastre:
I'd like to add support for "widthDp" and "heightDp" params

For that, I've tried the solution described in [this page](https://github.com/DeNA/android-modern-architecture-test-handson/blob/main/docs/handson/VisualRegressionTest_Preview_ComposablePreviewScanner.md#uimodeと画面サイズへの対応-myapplytorobolectricconfigurationの実装)
```kotlin
if (preview.previewInfo.widthDp != -1 && preview.previewInfo.heightDp != -1) {
        setDisplaySize(preview.previewInfo.widthDp, preview.previewInfo.heightDp)
}

private fun setDisplaySize(widthDp: Int, heightDp: Int) {
    val context = ApplicationProvider.getApplicationContext<Context>()
    val display = ShadowDisplay.getDefaultDisplay()
    val density = context.resources.displayMetrics.density
    widthDp.let {
        val widthPx = (widthDp * density).roundToInt()
        Shadows.shadowOf(display).setWidth(widthPx)
    }
    heightDp.let {
        val heightPx = (heightDp * density).roundToInt()
        Shadows.shadowOf(display).setHeight(heightPx)
    }
}
```

but it does not seem to work. Could be that I am missing something?

sergio-sastre:
So I got it working in a private project, it seems to work but one has to call `composeTestRule.activityRule.scenario.recreate()` afterwards.

I'll try to create a PR  to add this functionality to Roborazzi this week

sergio-sastre:
Seems 2 or 3 more adjustments were necessary to make the Views look exactly as in the Previews and avoid using a ComposeTestRule, which was making the test execution considerably slower.

One point to discuss will be the RoborazziTransparentTheme, which does
`android:windowActionBar = false` and that influences the size set via Robolectric qualifiers or the ShadowDisplay. It should be wither set to true or use another Activity.

I’ll create a PR today or tomorrow when I find time, which will also include support for “showBackground” and “backgroundColor”.

sergio-sastre:
@takahirom 

Draft PR #576

PR #576, opened 2024-11-24
Add support for heightDp, widthDp, showBackground, backgroundColor

sergio-sastre:
This addresses #559

Still very open for discussion


roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
```diff
@@ -0,0 +1,123 @@
...
+  widthDp: Int,
+  heightDp: Int,
+  showBackground: Boolean,
+  backgroundColor: Long,
+  filePath: String,
+  roborazziOptions: RoborazziOptions = provideRoborazziContext().options,
+  content: @Composable () -> Unit,
+) {
+  if (!roborazziOptions.taskType.isEnabled()) return
+  registerRoborazziActivityToRobolectricIfNeeded(ComponentActivity::class.java)
```
sergio-sastre, line 33: In my private project I had a custom Preview with heightDp = 30 and widthDp= 30 nad it crashed with the RobolectricTransparentActivity (actually, it would make sense to add tests for such cases)

That is because it has `android:windowActionBar = false`. The windowActionBar seems to be 56dp, so I assume that any height or width lower than that would make test crash. Moreover, I observed that Previews that set very high widthDp or heightDP values e.g. widthDp = 2000 and heightDp = 2000, had the height cut down by 56 dp...

That does not happen by directly using the ComponentActivity. Since I am not sure why the RobolectricTransparentActivity has `android:windowActionBar = false`, I am using ComponentActivity instead.

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
```diff
@@ -0,0 +1,123 @@
...
+
+      val viewRootForTest = composeView.getChildAt(0) as ViewRootForTest
+      viewRootForTest.view.captureRoboImage(filePath, roborazziOptions)
+    }
+
+    // Closing the activity is necessary to prevent memory leaks.
+    // If multiple captureRoboImage calls occur in a single test,
+    // they can lead to an activity leak.
+  }
+}
```
sergio-sastre, line 68: I am not sure where to put these methods `setBackgroundColor` and `setShadowDisplay`... unfortunately both are Activity dependent, so I cannot really put them together with the other ones inside `applyToRobolectricConfiguration()`

I am a bit concern about how to make users of that method aware that to successfully apply widthDp, heightDp, showBackground and backgroundColor, they need an Activity

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
```diff
@@ -0,0 +1,123 @@
...
+) {
+  when (showBackground) {
+    false -> window.decorView.setBackgroundColor(Color.TRANSPARENT)
+    true -> if (backgroundColor != 0L) {
+      window.decorView.setBackgroundColor(backgroundColor.toInt())
+    }
+  }
+}
+
+fun Activity.setShadowDisplay(
```
sergio-sastre, line 82: I think a better name for this would be `setDisplaySize`

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
takahirom: Thanks. I think it would be great if we could restore the color after the test. We can refer to @yschimke's implementation for this.

https://github.com/takahirom/roborazzi/pull/574/files#diff-dbee5836626b59835a193026779dc5c522d087620d2cbbad92ba72a15149ed1eR159

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
takahirom: Thank you for your information. We set windowActionBar to false because the activity has a shadow for the ActionBar. Do you know if the shadow is not shown with ComponentActivity?
https://github.com/takahirom/roborazzi/pull/283

sergio-sastre: I've just added the Preview with widthDp = 30 and heightDp = 30 and it is crashing although I am using the ComponentActivity diectly... in my project it does not crash and renders properly... it only crashes if I use the RobolectricTransparentActivity... I think it might be that I am using the latest version coming with androidx.compose:compose-bom:2024.11.00 and Roborazzi uses another one where it does not work?

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
```diff
@@ -117,7 +124,24 @@ private fun (@Composable () -> Unit).withSize(
...
+private fun registerRoborazziComposePreviewsActivityToRobolectricIfNeeded() {
+  try {
+    val appContext: Application = ApplicationProvider.getApplicationContext()
+    Shadows.shadowOf(appContext.packageManager).addActivityIfNotPresent(
+      ComponentName(
+        appContext.packageName,
+        ComponentActivity::class.java.name,
+      )
+    )
+  } catch (e: ClassNotFoundException) {
```
yschimke, line 143: I'm curious about this.

You are directly calling both robolectric and ComponentActivity. Wouldn't it fail with an error (NoClassDefFoundError , or similar) rather than a ClassNotFoundException?  I normally assume ClassNotFoundException is for reflection usage with a checked exception.

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
```diff
@@ -117,7 +124,24 @@ private fun (@Composable () -> Unit).withSize(
...
   return resizedPreview
+}
+
+private fun registerRoborazziComposePreviewsActivityToRobolectricIfNeeded() {
+  try {
+    val appContext: Application = ApplicationProvider.getApplicationContext()
+    Shadows.shadowOf(appContext.packageManager).addActivityIfNotPresent(
+      ComponentName(
+        appContext.packageName,
+        ComponentActivity::class.java.name,
```
yschimke, line 140: Should this be RoborazziComposePreviewsActivity?

sergio-sastre: I have not delete the RoborazziCompoaePreviewsActivity because I’m still experimenting😅.
Anyway, I still get the same result with both unfortunately… seems the height is not exactly what it should by some pixels 🧐

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
sergio-sastre: I agree and haven’t realised that since I copy pasted it from the analogue method in RoborazziCompose.kt file.

worth checking 👍

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
sergio-sastre: Since the activity is closed after snapshoting, it should not be necessary

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziComposePreview.kt
takahirom: Thanks. Do you have a test we can use to reproduce the issue? Is the failing test in this PR related to that problem?

takahirom: I'm not sure, but the hieght has become zero. This is only my guess, but:
1. Setting Activity.setDisplaySize
2. Set action bar size to 83
3. Set ViewGroup padding to 83
4. The size for layout become zero


image
image
image
image

takahirom: I think you can debug this issue by adding breakpoint on AbstractComposeView.internalOnMeasure

takahirom: I took a quick look, and it seems that the `android:windowActionBar` isn't applied because we can't merge themes in tests. The ActionBar was created, and it takes a margin. I've switched to using system `Theme_Translucent_NoTitleBar_Fullscreen`, and it works. I'm not sure if there is a better theme available, but this is a step forward.

image

https://github.com/takahirom/roborazzi/pull/576/commits/b4ff8cfab7e5460607569b24efd15260bdd41e62

sergio-sastre: Awesome! Once that works, I’ll try to clean this up and create a non-draft PR when I find the time.

thanks! I highly appreciate it 🙏

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -14,10 +14,25 @@ fun ComposablePreview<AndroidPreviewInfo>.captureRoboImage(
 ) {
   val composablePreview = this
   composablePreview.applyToRobolectricConfiguration()
-  captureRoboImage(filePath = filePath, roborazziOptions = roborazziOptions) {
-    composablePreview()
-  }
+  captureRoboImage(
+    filePath = filePath,
+    roborazziOptions = roborazziOptions,
+    content = { activityScenario ->
```
sergio-sastre, line 20: I think it was practical to add such a captureRoboImage method to allow users to set properties through the activityScenario. Open to other opinions

sergio-sastre (commented):
@takahirom 
Feel free to review in detail. It's likely you'd like to do something differently to allow users of the api to set the background and dimensions via the PreviewTester?

sergio-sastre:
I’m a bit unsure why the builds fail?

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziCompose.kt
```diff
@@ -48,20 +42,55 @@ fun captureRoboImage(
...
-    Shadows.shadowOf(appContext.packageManager).addActivityIfNotPresent(
-      ComponentName(
-        appContext.packageName,
-        RoborazziTransparentActivity::class.java.name,
-      )
-    )
-  } catch (e: ClassNotFoundException) {
-    // Configured to run even without Robolectric
-    e.printStackTrace()
+fun captureRoboImage(
```
takahirom, line 45: I'm not sure, but I think `captureRoboImages` might have the same signature in the JVM bytecode. Could you check it out?

> e: file:///Users/runner/work/roborazzi/roborazzi/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt:155:5 Overload resolution ambiguity:

takahirom:
I noticed that there was a shadow at the top in the comparison images, but it has now been removed. It looks great! I’ll review the changes tomorrow.

sample-generate-preview-tests/src/main/java/com/github/takahirom/preview/tests/Previews.kt
```diff
@@ -99,4 +99,107 @@ fun PreviewWithProperties2() {
...
+  ) {
+    Text(
+      modifier = Modifier.padding(8.dp),
+      text = "Hello, World!"
+    )
+  }
+}
+
+@Preview(
+  name = "Preview width & height",
```
sergio-sastre, line 122: "Preview width & height small"?

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
takahirom: Thanks. Designing these APIs is a little difficult. I'm going to give a talk at a technology conference this Saturday. Sorry for being late, but I would like to consider this over the weekend.

takahirom:
@sergio-sastre 
This might be over-engineering, but I made a change to make it pluggable. With this change, we can provide the same flexibility to `captureRoboImage(composable)` as the Preview functionality, without needing `captureRoboImageWithActivityScenarioSetup()`. Could you take a look?
https://github.com/sergio-sastre/roborazzi/compare/feature/support_previews_heightDp_widthDp_showBackgorund_backgroundColor...takahirom:roborazzi:takahirom/refactor-plugable/2024-12-01?diff=split&w=

sergio-sastre:
> @sergio-sastre Thank you for your patience while waiting. This might be over-engineering, but I made a change to make it pluggable. With this change, we can provide the same flexibility to `captureRoboImage(composable)` as the Preview functionality, without needing `captureRoboImageWithActivityScenarioSetup()`. Could you take a look? https://github.com/sergio-sastre/roborazzi/compare/feature/support_previews_heightDp_widthDp_showBackgorund_backgroundColor...takahirom:roborazzi:takahirom/refactor-plugable/2024-12-01?diff=split&w=

I’ll do it either today or tomorrow when I find time😊

PR #577, opened 2024-12-01
Refactor  Add support for heightDp, widthDp, showBackground, backgroundColor

takahirom:


roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -9,15 +9,26 @@ import sergio.sastre.composable.preview.scanner.core.preview.ComposablePreview
 
 @ExperimentalRoborazziApi
 fun ComposablePreview<AndroidPreviewInfo>.captureRoboImage(
-  filePath: String = DefaultFileNameGenerator.generateFilePath(),
-  roborazziOptions: RoborazziOptions = RoborazziOptions(),
+  filePath: String,
+  roborazziOptions: RoborazziOptions = provideRoborazziContext().options,
+  applierBuilder: RoborazziComposeApplierBuilder = RoborazziComposeApplierBuilder()
```
sergio-sastre, line 14: it is true that it adds some complexity, but brings the benefit of making the api more flexible.

Therefore, I have a couple of points:
1. We are inconsistent in the way we handle fontSize, UiMode, Locale, device...etc. vs.  sized and colored.
I think it would become much more intuitive if all of them would be handled the same way, e.g.

```kotlin
RoborazziComposeApplierBuilder()
   .sized(
       widthDp = previewInfo.widthDp,
       heightDp = previewInfo.heightDp
     )
     // I'd rather call this background than coloured. Both properties include background inn their names indeed. One could also call them : `show` instead of `showBackground` and `color` instead of `backgroundColor`
     .background( 
       showBackground = previewInfo.showBackground,
       backgroundColor = previewInfo.backgroundColor
     )
     .uiMode(
        configurationUiMode  = previewInfo.uiMode
     )
     .locale(
        bcp47LanguageTag = previewInfo.locale
     )
     ...
```

I think symmetry in code also makes it more understandable

2.  The name of this variable is a bit ambiguous. I'm not sure what to expect from this as a user of this method. I think something in the direction of `previewOptions` and `RoborazziPreviewOptionsBuilder` or `composableConfig`and `RoborazziComposableConfigBuilder` is more understandable

sergio-sastre (commented):
@takahirom 
I just left a couple of general remarks.

Feel free to address them if you find them appropriate

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
takahirom: Thank you for your review. That makes sense.  
I've migrated the settings such as local to the applier and renamed the applier to config.
https://github.com/takahirom/roborazzi/pull/577/commits/a899893ff06c10fd53fa6a5d76bffe0eb01b8d49
https://github.com/takahirom/roborazzi/pull/579/files

sergio-sastre: Looks good to me. 
The only improvements I’d propose are input validations (e.g. fontScale should not be lower than 0 and the like) and logging messages for invalid inputs.

apart from that feel free to merge 😊

PR #588, opened 2024-12-05
add localInspectionMode property to RoborazziComposeOptions

YusukeMoriJapan:
fix this issue #589

### DIFF
--- a/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
+++ b/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
@@ -178,9 +178,16 @@ class ManualTest {
       roborazziComposeOptions = RoborazziComposeOptions {
         // We have several options to configure the test environment.
         fontScale(2f)
-        /* The default value is false, but we can set it to true,
-        if you want to use the logic for Preview in composable functions. */
-        localInspectionMode(false)
+
+        /*
+        We don't specify `localInspectionMode` by default.
+        The default value for `localInspectionMode` in Compose is `false`.
+        This is to maintain higher fidelity in tests.
+        If you encounter issues integrating the library, you can set `localInspectionMode` to `true`.
+
+        localInspectionMode(true)
+         */
+
         // We can also configure the activity scenario and the composable content.
         addOption(
           object : RoborazziComposeComposableOption,
--- a/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
+++ b/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
@@ -40,6 +40,7 @@ import com.github.takahirom.roborazzi.captureRoboGif
 import com.github.takahirom.roborazzi.captureRoboImage
 import com.github.takahirom.roborazzi.captureRoboLastImage
 import com.github.takahirom.roborazzi.fontScale
+import com.github.takahirom.roborazzi.localInspectionMode
 import com.github.takahirom.roborazzi.roboOutputName
 import com.github.takahirom.roborazzi.roborazziSystemPropertyOutputDirectory
 import com.github.takahirom.roborazzi.withComposeTestTag
@@ -177,6 +178,9 @@ class ManualTest {
       roborazziComposeOptions = RoborazziComposeOptions {
         // We have several options to configure the test environment.
         fontScale(2f)
+        /* The default value is false, but we can set it to true,
+        if you want to use the logic for Preview in composable functions. */
+        localInspectionMode(false)
         // We can also configure the activity scenario and the composable content.
         addOption(
           object : RoborazziComposeComposableOption,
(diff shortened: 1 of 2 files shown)


roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -18,7 +20,9 @@ fun ComposablePreview<AndroidPreviewInfo>.captureRoboImage(
   if (!roborazziOptions.taskType.isEnabled()) return
   val composablePreview = this
   captureRoboImage(filePath, roborazziOptions, configBuilder) {
-    composablePreview()
+    CompositionLocalProvider(LocalInspectionMode provides true) {
+      composablePreview()
+    }
   }
```
YusukeMoriJapan, line 26: When running the Preview Screenshot Test, it is normally expected to produce the same result as during the Preview rendering, so there is no option to choose between true or false. If there is a need to make it selectable, I will modify it accordingly.

takahirom:
Thanks. I think there is a trade-off when we use `LocalInspectionMode provides true`. It is important to maintain fidelity (how close it is to production) in tests, and if we use it, the code we run will be altered. Therefore, I don't want to make it the default option. How about adding a `RoborazziComposeConfigBuilder` that enables and disables the inspection mode so users can configure it?

YusukeMoriJapan:
> Thanks. I think there is a trade-off when we use `LocalInspectionMode provides true`. It is important to maintain fidelity (how close it is to production) in tests, and if we use it, the code we run will be altered. Therefore, I don't want to make it the default option. How about adding a `RoborazziComposeConfig` that enables and disables the inspection mode so users can configure it?

OK👍
I will try it now.

takahirom:
Sorry, I intended to add the configuration to `RoborazziComposeConfig` instead of `RoborazziOptions`. RoborazziOptions is a option for all images not for Compose. `RoborazziComposeConfig` is [a new feature](https://github.com/takahirom/roborazzi/pull/590#issuecomment-2519777276) that hasn’t been released yet, and the API is still under consideration. 

I suggest rebasing this PR onto the `12-05-refactor_roborazzicomposeconfigbuilder` branch to use the new `RoborazziComposeConfig.Builder`.

YusukeMoriJapan:
> Sorry, I intended to add the configuration to `RoborazziComposeConfig` instead of `RoborazziOptions`. RoborazziOptions is a option for all images not for Compose. `RoborazziComposeConfig` is [a new feature](https://github.com/takahirom/roborazzi/pull/590#issuecomment-2519777276) that hasn’t been released yet, and the API is still under consideration.
> 
> I suggest rebasing this PR onto the `12-05-refactor_roborazzicomposeconfigbuilder` branch to use the new `RoborazziComposeConfig.Builder`.


I just realized the same thing. Thank you for letting me know.
I will rebase and try again.

takahirom:
Sorry, I'm going to change the APIs a little bit more. The timing is a bit unfortunate 🙇. I suggest you try this tomorrow.

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
takahirom: We could have something like `roborazz.compose.inspectionmode = true`. The default would be null. However, this might be a little confusing because we can't inject the setting if you use your activity.

sergio-sastre: Then better to leave it as ComposeOption

takahirom:
Thank you for your patience. I believe you can work on this now.

sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ManualTest.kt
```diff
@@ -177,6 +178,9 @@ class ManualTest {
       roborazziComposeOptions = RoborazziComposeOptions {
         // We have several options to configure the test environment.
         fontScale(2f)
+        /* The default value is false, but we can set it to true,
+        if you want to use the logic for Preview in composable functions. */
+        localInspectionMode(false)
```
takahirom, line 183: I don't think we need to apply localInspectionMode explicitly. How about making it a comment?

```suggestion
        /*
        We don't specify `localInspectionMode` by default. The default value for `localInspectionMode` in Compose is `false`. This is to maintain higher fidelity in tests.
        If you encounter issues integrating the library, you can set `localInspectionMode` to `true`.
        localInspectionMode(true)
        */

YusukeMoriJapan: @takahirom 
OK, I have incorporated your feedback into the comment.
https://github.com/takahirom/roborazzi/pull/588/commits/4f6007ad6296d48abf4c264b7d6ccc097176dc7c

takahirom (approved):
Looks great! Thank you for your contribution!

PR #600, opened 2024-12-11
Sort units and actions of ComposePrintToString to stabilize output in Dump screenshots.

siarhei-luskanau:
Dump screenshot has node properties. Sometime Roborazzi generates new screenshots (or fails verification) due the node properties order. This Pull Request add the output sorting to avoid the issue.

Screenshot Dump changes


takahirom:
Thank you for pull request. Could you check the CI build?
image

PR #604, opened 2024-12-12
Add overlap workaround

takahirom:


include-build/roborazzi-core/src/androidMain/kotlin/com/github/takahirom/roborazzi/ComposeScreenshot.kt
```diff
@@ -56,7 +57,40 @@ fun SemanticsNode.fetchImage(recordOptions: RoborazziOptions.RecordOptions): Bit
...
+    windowToUse.hasFeature(Window.FEATURE_ACTION_BAR) &&
+    actionBarWorkaroundIsOn
+  if (shouldUseComposeCapture) {
+    roborazziReportLog(
+      "Using ComposeView.draw to capture the bitmap to avoid content overlap issues with the ActionBar.\n" +
+        "Window-based capture (which typically provides higher fidelity) can cause these overlap issues when an ActionBar is present.\n" +
+        "We recommend setting the theme using <application android:theme=\"@android:style/Theme.Material.NoActionBar\" /> in your test/AndroidManifest.xml to fix this.\n" +
+        "If you are intentionally using an ActionBar, you can disable this workaround by setting the gradle property 'roborazzi.compose.actionbar.overlap.fix' to false.\n" +
+        "This problem is tracked in https://issuetracker.google.com/issues/383368165"
+    )
```
sergio-sastre, line 83: I believe this could be the issue with the missing shadows. It is likely this does not use pixel copy under the hood to draw the bitmap? In that case elevation is ignored (e.g. shadows)

A solution would be to use PixelCopy when generating the bitmap. You can check an implementation here
https://github.com/sergio-sastre/AndroidUiTestingUtils/blob/349840245adf9ff8a4e20be6467f6d30e15b3aa2/utils/src/main/java/sergio/sastre/uitesting/utils/utils/DrawToBitmapExt.kt#L127

takahirom: @sergio-sastre 

Thank you for acting as a consultant on this issue.  
I think we can't use the `PixelCopy` approach or the approach of hiding the ActionBar. I made a summary—could you take a look? There might be a better solution.

## Roborazzi Main Branch Behavior (Using PixelCopy)

| SDK | `createComposeRule()` and `onRoot().captureRoboImage()` Behavior | Image |
|---|---|---|
| 34 or below | Screenshots can be taken with the ActionBar shadow.  |  image |
| 35 or higher | **Screenshots are overlapped by the ActionBar because SDK 35 doesn't provide padding for the ActionBar (I believe this is a major degradation, and [I filed an issue for this](https://issuetracker.google.com/issues/383368165). However, it may take a long time to fix.)**   | image |

## If We Use This Branch's Fix (Using `view.draw()` Instead of PixelCopy)

The reason for using `view.draw()` is to fix the issue of the overlapping ActionBar, as capturing the window via PixelCopy includes the overlapping ActionBar. 
Additionally, as we use `onRoot().captureRoboImage()`, we can't use `RoborazziTransparentActivity` in this case.

This workaround is applied only when there is an ActionBar and the SDK version is 35.

| SDK | `createComposeRule()` and `onRoot().captureRoboImage()` Behavior | Image |
|---|---|---|
| 34 or below | Screenshots can be taken with the ActionBar shadow (**No change**).  |  image |
| 35 or higher | **Screenshots are no longer overlapped by the ActionBar, but the shadow is not drawn, which is a degradation.** (A warning log has been added for this.)   | image |

## What About Turning Off the ActionBar Automatically When Calling `captureRoboImage`?

I initially tried this approach, but I identified some trade-offs:

* Changing the app layout might cause layout invalidation and recalculation.
* Test isolation should be maintained (although we could revert the ActionBar after calling `captureRoboImage`, this might still cause additional layout invalidation).
* What if we want to use the ActionBar? (This might not necessarily need to be considered.)

This workaround also introduces significant trade-offs.

sergio-sastre: > @sergio-sastre
> 
> Thank you for acting as a consultant on this issue. I think we can't use the `PixelCopy` approach or the approach of hiding the ActionBar. I made a summary—could you take a look? There might be a better solution.
> 
> ## Roborazzi Main Branch Behavior (Using PixelCopy)
> SDK	`createComposeRule()` and `onRoot().captureRoboImage()` Behavior	Image
> 34 or below	Screenshots can be taken with the ActionBar shadow.	image
> 35 or higher	**Screenshots are overlapped by the ActionBar because SDK 35 doesn't provide padding for the ActionBar (I believe this is a major degradation, and [I filed an issue for this](https://issuetracker.google.com/issues/383368165). However, it may take a long time to fix.)**	image
> ## If We Use This Branch's Fix (Using `view.draw()` Instead of PixelCopy)
> The reason for using `view.draw()` is to fix the issue of the overlapping ActionBar, as capturing the window via PixelCopy includes the overlapping ActionBar. Additionally, as we use `createComposeRule()`, we can't use `RoborazziTransparentActivity` in this case.
> 
> This workaround is applied only when there is an ActionBar and the SDK version is 35.
> 
> SDK	`createComposeRule()` and `onRoot().captureRoboImage()` Behavior	Image
> 34 or below	Screenshots can be taken with the ActionBar shadow (**No change**).	image
> 35 or higher	**Screenshots are no longer overlapped by the ActionBar, but the shadow is not drawn, which is a degradation.** (A warning log has been added for this.)	image
> ## What About Turning Off the ActionBar Automatically When Calling `captureRoboImage`?
> I initially tried this approach, but I identified some trade-offs:
> 
> * Changing the app layout might cause layout invalidation and recalculation.
> * Test isolation should be maintained (although we could revert the ActionBar after calling `captureRoboImage`, this might still cause additional layout invalidation).
> * What if we want to use the ActionBar? (This might not necessarily need to be considered.)
> 
> This workaround also introduces significant trade-offs.

Thanks for taking the time to explain the consequences of each approach! That was really valuable.

so now I understand why PixelCopy would not work.

However, regarding hiding the action bar, I am not very convinced of the trade-offs.

1. Without real measures it is hard to confirm, but I think layout recalculation in this case should not be expensive, even if noticeable. What is true is that one should wait for that to recalculation to finish before snapshoting and that can be achieved with `InstrumentationRegistry.getInstrumentation().waitForIdleSync()` which works with Robolectric as far as I know.
2. TestIsolation is guaranteed by the ComposeTestRule. It closes the activity that was launched, so every test launches a new activity where the action bar is initially shown.
3. I think this is irrelevant since this is actually a workaround to avoid that.

In my personal opinion, I’d rather go for hiding the action bar.

that would be consistent among all versions, I think the impact regarding test execution is not gonna be very noticeable, and the workaround would be just temporary.

Just my 2 cents. It is up to you how to workaround this issue with sdk 35.

I am also wondering… if the problem lies on ComposeTestRule on SDK 35 (not sure if that is the issue), maybe is worth downgrading the library version to see if the bug was introduced in one of the last versions?

takahirom: Thank you for your feedback.

Setting the theme and requesting features do not work after setting the content view. Therefore, I used `view.getActivity()?.actionBar?.hide()`.

I hide the ActionBar but revert it to its original state after capturing the image. I believe this is necessary to avoid affecting other test checks, as `captureRoboImage` should not change anything after the capture.

PR #611, opened 2024-12-16
Sort key order of ComposePrintToString to stabilize output in Dump sцreenshots.

siarhei-luskanau:
Dump screenshot has node properties. Sometime Roborazzi generates new screenshots (or fails verification) due the node properties order. This Pull Request add the output sorting to avoid the issue.


takahirom (approved):
Thank you for your contribution!🔑

Issue #615, opened 2024-12-20
Previously recorded screenshots keep reappearing


pedromfmachado:
# Description
We've recently introduced Roborazzi to our project and are currently in the phase of migrating/writing new Roborazzi tests. When doing so, we faced a weird issue where previously recorded screenshots keep reappearing. In sum, here's a short summary of the flow to reproduce the issue:
- Run `./gradlew recordRoborazziDebug`
- Delete the recorded files
- Change the test so the resulting files are different
- Run `./gradlew recordRoborazziDebug` again
- Both the new files and the old files would appear

The interesting part is that this happens even when `build/intermediates/roborazzi` is deleted, leading me to believe this is rather related to how the outputs are cached.

# Possible root cause
When trying to understand the problem, I stumbled on [this portion of code](https://github.com/takahirom/roborazzi/blob/64ab45f344dd28bb09b8a31cb2e9bf08ba41a8fd/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt#L479):
```Kotlin
// Copy all files from outputDir to intermediateDir
// so that we can use Gradle's output caching
test.infoln("Roborazzi: test.doLast Copy files from ${outputDir.get()} to ${intermediateDir.get()}")
// outputDir.get().asFileTree.forEach {
//   println("Copy file ${finalizeTask.absolutePath} to ${intermediateDir.get()}")
// }
outputDir.get().asFile.mkdirs()
outputDir.get().asFile.copyRecursively(
  target = intermediateDir.get().asFile,
  overwrite = true
)
```

If my understanding is correct, this means that the outputs are copied into the `intermediates/roborazzi` directory together with anything that was already in there previously, leading to the behaviour I described before.

# Possible solution
The contents of the `intermediates/roboazzi` folder are being cleared within the [`cleanupOldScreenshotsIfNeeded` method](https://github.com/takahirom/roborazzi/blob/64ab45f344dd28bb09b8a31cb2e9bf08ba41a8fd/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt#L619), but only when `cleanupOldScreenshots` is set to true:

```Kotlin
// Delete all images from the intermediateDir
intermediateDir.get().asFile.walkTopDown().forEach { file ->
  if (KnownImageFileExtensions.contains(file.extension)) {
    file.delete()
  }
}
```

[...]

PR #616, opened 2024-12-20
Removing files from intermediate dir to prevent old screenshots from reappearing

pedromfmachado:
Proposed solution for #615

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
@@ -132,25 +132,6 @@ class RoborazziGradleProjectTest {
   }
 
   @Test
-  fun recordWhenRunTwiceAfterTestChanges() {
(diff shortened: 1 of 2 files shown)


takahirom:
Thanks! We need a test to ensure this type of error is caught. Could you add one that specifically triggers this failure?

pedromfmachado:
I've added the test and confirmed it was triggering the issue before the fix was applied. It also helped me improve the fix to use `test.systemProperties` instead of `roborazziProperties` as the previous solution was only working correctly when the system parameter (`-Proborazzi.test.record=true`) was used.

takahirom:
l'm concerned that this implementation might remove other reference (golden) images when users save images in GitHub and run tests with the --tests className .method option or category filters to update screenshots. It seems like we need a test case to cover this scenario. Do you have time to handle this?

pedromfmachado:
I've added a test to cover the filtering case, it looks like the implementation doesn't cause any problems with that.
I was, however, a bit confused with the reference to `src/screeeshots` in your comment, did you mean that it's also relevant to add two similar tests for the custom output directory use case?

takahirom:
This is quite difficult to handle correctly, and it's very challenging. I've checked several patterns locally, and this appears to work as intended. 
Thank you for your excellent contribution!

PR #618, opened 2024-12-22
Improved roborazzi.generateComposePreviewRobolectricTests.packages specification

TBSten:
`roborazzi.generateComposePreviewRobolectricTests.packages` to set `android.namespace` instead of immediately raising an error if it is not set.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-VERSION_NAME=1.59.0
+VERSION_NAME=1.60.0
 GROUP=io.github.takahirom.roborazzi
 # Project-wide Gradle settings.
 


takahirom:
Thanks. I think this change has a minor tradeoff. When users can't take screenshots because of the package name, it can be difficult to troubleshoot the issue. However, if we require setting the package name, users will understand why screenshots are disabled. 

What are your thoughts on this problem? It's similar to the CompositionLocal vs. parameter dilemma in Compose.

TBSten:
As you say, it would be better not to set default values for packages on the library side (roborazzi), since there are problems depending on the environment of the project...
It would be better to use a common build configuration within the project.

However, when I try to set default values for roborazzi.packages using composite build in my project, I can't set default values because of a check by the roborazzi gradle plugin.

```kt
// roborazzi.packages is checked first and a configuration error occurs.
afterEvaluate {
  if(packages.get().orEmpty().isEmpty()) {
    packages = listOf(androidExtension.namespace)
  }
}
```

Is it possible to delay the check until the task is performed as my updated Pull Request indicates to address this issue?

takahirom:
Thanks. I think we can set it like `packages.set(provider { listOf(android.namespace) })`. Could you check it out?

TBSten:
Oh! I was able to set it up successfully! Thank you...!

I wonder when is the best time for Gradle Plugin validation... 🤔 (some things are not wrapped in Property and may get stuck)

TBSten:
I leave it up to you to decide if you want to merge this PR or not.

Thank you for your response!

PR #622, opened 2024-12-27
Sort capture results to achieve deterministic order in HTML report

### LINKED ISSUE #621: Order of capture results is not deterministic in HTML report
While working on a screenshot test battery with many parameterized variations of the same screen, I noticed that the HTML report does not show its results in alphabetical order. For example, when running `sample-android:recordRoborazziDebug` and inspecting the report, the first element related to `WindowCaptureTest`, then `ComposeTest`, then `Sdk25DontWorkScreenshotTest` and so on – see screenshot below. It would be nice to group related previews together so they appear next to each other.

Screenshot 2024-12-27 at 12 09 29

I have prepared a fix for this and will open a PR momentarily if it suits you.

mannodermaus:
Sort capture results by the file path of their golden image, achieving a deterministic order of screenshots that groups together related images.

|Before|After|
|---|---|
|Screenshot 2024-12-27 at 12 09 29|Screenshot 2024-12-27 at 12 10 47|

Resolves #621.


takahirom:

PR #625, opened 2025-01-03
Show diff when size changed

### LINKED ISSUE #623: Diff screenshots should be as large as the largest of the two compared screenshots
Hello, first, thank you for this framework, it works great. :)

I'm not sure if this is a bug, but it feels like it; I certainly think making this change would improve the diffs.

When comparing two components with different sizes, the diff screenshot takes the size of the reference. Therefore, if the new component got taller or wider, the diff will not correctly show the changed pixels.

I think making it so that the diff's dimensions are always large enough to accomodate the reference and the new version would solve this issue.

For example:

CallToActionTests SecondaryDefault Light rtlnl_compare

I would expect the full `sit amet,` to be visible in red here.

takahirom:
close #623


takahirom:
image
It seems work well

Issue #233, opened 2023-12-29
Tests succeeding locally but failing in the build pipeline


beshoy-samy:
Hi guys I am facing a weird issue where the tests are succeeding locally but failing in the build pipeline
here is one of the compared images.

It seems there is no difference between the reference and the new one, so what could be the issue here? 🤔 

I am using the latest version `1.9.0-alpha-2`

<img src="https://github.com/takahirom/roborazzi/assets/16430789/02bc4b4b-2ee2-4dfc-94b1-78767c2ce5fe" />

takahirom:
@beshoy-samy 
Thank you for reporting this issue. Do you take reference screenshots in different environments? This is important because it appears that Robolectric Native Graphics renders differently across various operating systems.
https://github.com/robolectric/robolectric/issues/8457

takahirom:
Roborazzi offers two options for comparing images:

Percentage of Changed Pixels
Percentage of Color Difference
You can set these in the 'Compare Options' menu.
If you would like to accept these changes, please select these options.
https://github.com/takahirom/roborazzi/blob/main/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/RoborazziOptions.kt#L164

beshoy-samy:
@takahirom yep thanks a lot I did go around it using these options

PR #633, opened 2025-01-18
Support DelayedPreviews (not tested, just initial push)

sergio-sastre:


takahirom:
👀

takahirom:
The build failed because of a setting in the root build.gradle file:

```
tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
  compilerOptions {
    freeCompilerArgs.addAll(
      "-Xopt-in=com.github.takahirom.roborazzi.InternalRoborazziApi",
      "-Xopt-in=com.github.takahirom.roborazzi.ExperimentalRoborazziApi",
    )
  }
}
```

I've addressed this by excluding the "roborazzi-annotations" module from this setting.

takahirom:
I understand this is difficult, especially regarding ComposeRule. I'm exploring solutions. Please bear with me.

takahirom:
@sergio-sastre 
Thank you for your contribution! I've made some minor changes to utilize RoborazziComposeOptions, which I believe results in a slight improvement. However, we may need to consider the preview changes. 👀
https://github.com/takahirom/roborazzi/pull/633/files/4c4646c6e01373c1becf9bfc9c49b4c52deb0499..2764156a105defb07e7b21376de596e2d46f08d6

takahirom:
I think we might have multiple activities for each capture. I'm looking a little further into this.

sergio-sastre:
Great work!
Let me know your thoughts after exploring and we can discuss how to better solve the problem 😊

takahirom:
Thanks! I'm considering whether the name "DelayedPreview" is appropriate.  
Currently, using this implementation prevents us from utilizing the `activityTheme()` option, which isn't used by the PreviewScreenshots parameter. While this shouldn't pose an immediate issue, it limits the robustness of the RoborazziComposeActivityScenarioCreatorOption API. I'm open to suggestions for improvement but currently lack a concrete solution.

sergio-sastre:
> Thanks! I'm considering whether the name "DelayedPreview" is appropriate.
> Currently, using this implementation prevents us from utilizing the `activityTheme()` option, which isn't used by the PreviewScreenshots parameter. While this shouldn't pose an immediate issue, it limits the robustness of the RoborazziComposeActivityScenarioCreatorOption API. I'm open to suggestions for improvement but currently lack a concrete solution.

I've pushed a preview with the DelayedPreview annotation and the current implementation does not seem to work.
I'd say, let's focus first on having a working implementation, and then see what works or doesn't work and search for possible solutions

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -143,18 +197,28 @@ interface ComposePreviewTester<T : Any> {
 
 @ExperimentalRoborazziApi
 class AndroidComposePreviewTester : ComposePreviewTester<AndroidPreviewInfo> {
+  private val composeTestRule by lazy { createAndroidComposeRule<RoborazziActivity>() }
```
sergio-sastre, line 200: @takahirom
Actually, it does not matter whether lazy or not since it is used in EVERY test.

I was afraid that using this TestRule in every test would increase the execution time, but I haven't noticed any significant increase though... I believe that is different from Activity Scenario, which launches an Activity and that does add execution time indeed...

It could also be that my laptop (Mac M2) is very fast and the amount of preview tests is not that high (less than 15...). I'm unsure about this

roborazzi-compose/src/main/java/com/github/takahirom/roborazzi/RoborazziCompose.kt
```diff
@@ -118,8 +131,10 @@ private fun ActivityScenario<out ComponentActivity>.captureRoboImage(
         val composeView = activity.window.decorView
           .findViewById<ViewGroup>(android.R.id.content)
           .getChildAt(0) as ComposeView
+
         @SuppressLint("VisibleForTests")
         val viewRootForTest = composeView.getChildAt(0) as ViewRootForTest
+        doBeforeCapture()
```
sergio-sastre, line 137: @takahirom 
I'm open to other solutions, but it does have to happen at this point. That was the reason why it was not working previously: the `advanceTimeBy()` was being before the Composable was set as content in the Activity

roborazzi-annotations/src/main/java/com/github/takahirom/roborazzi/annotations/DelayCaptureRoboImage.kt
```diff
@@ -0,0 +1,6 @@
+package com.github.takahirom.roborazzi.annotations
+
+// TODO -> maybe add also parameter for ignoreFrames, as used in mainClock.advanceTime()
+// TODO -> Make BINARY annotation only applicable to Methods
+// TODO -> Docu: mention about the 16ms frame in Android
+annotation class DelayCaptureRoboImage(val delayInMillis: Long)
```
sergio-sastre, line 6: @takahirom 
Feel free to give feedback on the name ;)

I'm thinking whether it'd also make sense to add the capability to take screenshots after a set of time intervals... what could be interesting to verify animations.

I think in that case, we could create a different annotation for that like
`IntervalCaptureRoboImage(intervalsInMillis: arrayOf(500, 100, 200))`

If you like the idea, we could make it either in this MR or afterwards, up to you :)

takahirom: I think the name `CaptureRoboImage` is a little misleading when used for `DelayCaptureRoboImage`, because we filter Previews by `@Preview` and not by `@DelayCaptureRoboImage`. So I've renamed it to `RoboManualAdvance` .

As for `IntervalCaptureRoboImage`, could we perhaps have a repeatable annotation?
When considering `IntervalCaptureRoboImage`, the use of `CaptureRoboImage` name is understandable though.

```
@RoboManualAdvance(advanceTimeMillis = 100)
@RoboManualAdvance(advanceTimeMillis = 200)
@RoboManualAdvance(advanceTimeMillis = 500)
```

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
takahirom: I believe test isolation is important. I think it's good to have a separate Activity for each test, as we do now. Of course, it depends on how slow it is, but I'm fine with it.

roborazzi-annotations/src/main/java/com/github/takahirom/roborazzi/annotations/DelayCaptureRoboImage.kt
sergio-sastre: Thanks for the suggestion!
I am also not very convinced about RoboManualAdvance, I think it does not tell its purpose clearly.

Its purpose is to advance the clock before capturing a screenshot.
I think something similar to
`AdvanceClockBeforeCapture` or the like makes more sense.

as of repeatable annotations, ComposablePreviewScanner still does not support them as stated in the corresponding method:

https://github.com/sergio-sastre/ComposablePreviewScanner/blob/ea9962eada536d81b9c8cdfa3df586df03c182a9/core/src/main/java/sergio/sastre/composable/preview/scanner/core/preview/ComposablePreview.kt#L34

That lies on me since it is a bit tricky 😅.
However, I agree that is a better solution.
I can create an issue and try to address it to support that in the future

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
sergio-sastre: I believe that too. However, even with this lazy, the ComposeTestRule is still being created in EVERY test, because it is called in other parts of the AndroidPreviewTester ( I’ve debugged it).
I’d also like it to be created when necessary … any ideas on how to avoid that?

roborazzi-annotations/src/main/java/com/github/takahirom/roborazzi/annotations/DelayCaptureRoboImage.kt
sergio-sastre: I‘m planning to release support for repeated annotations in ComposablePreviewScanner 0.5.1.

However, what would be your expectation?
Would  „duplicate“ previews for each repeated annotation, or would be more like a `getRepeatedAnnotation()` and get a list with all annotations?
I‘m more inclined to the last one, but I‘m wondering how we could integrate it in this case 🤔

takahirom:
```
@RoboManualAdvance(advanceTimeMillis = 100)
@RoboManualAdvance(advanceTimeMillis = 200)
@RoboFoo(true)
@RoboFoo(false)
```

We might want to take a screenshot when RoboFoo is true and RoboManualAdvance is 100. So it might be good to have one annotation like `@RoboConfig()`?

takahirom: I have renamed RoboManualAdvance to RoboComposePreviewOptions.

sergio-sastre: @takahirom 
So, I’ve been trying to support repeatable annotations. For “main” sourceSet it is fine, but for other sourcesets, like screenshotTest and androidTest it is quite challenging.
But that is because for those sourcesSets ComposablePreviewScanner uses the corresponding compiled Classes, and there Repeatable annotations are compiled differently (that was also a challenge with Previews 😅).
Supporting this would be add some complexity to the code and I’m unsure if it’s worth the hassle

so there are 2 ways to go:
1. We find another way to support this case without repeatable annotations, e.g. passing an array as argument with the delays
2. I release experimental support for repeatable annotations, working on “main” for now.

what are your thoughts?

sergio-sastre: @takahirom 
So, I’ve been trying to support repeatable annotations. For “main” sourceSet it is fine, but for other sourcesets, like screenshotTest and androidTest it is quite challenging.
But that is because for those sourcesSets ComposablePreviewScanner uses the corresponding compiled Classes, and there Repeatable annotations are compiled differently (that was also a challenge with Previews 😅).

so there are 2 ways to go:
1. We find another way to support this case without repeatable annotations, e.g. passing an array as argument with the delays
2. I release 

> Anyway, I have renamed RoboManualAdvance to RoboComposePreviewOptions.

Honestly, I think that would be a separate thing.
This is about taming screenshots at different times.
RoboComposePreviewOptions sounds like a way to pass ComposePreviewOptions via the annotation.
I think it is better to keep things separate, to avoid the users of the api get confused and also make clearer separation of functionalities in code.

sergio-sastre: > Anyway, I have renamed RoboManualAdvance to RoboComposePreviewOptions.

Honestly, I think that would be a separate thing.
This is about taming screenshots at different times.
RoboComposePreviewOptions sounds like a way to pass ComposePreviewOptions via the annotation.
I think it is better to keep things separate, to avoid the users of the api get confused and also make clearer separation of functionalities in code.

takahirom: Thank you for your opinion.  
Regarding annotation separation, I initially thought it would be challenging to support both **variation screenshots** and **representative patterns**. Do you have any ideas to facilitate this? (The theme serves as an example of another annotation type)

**Use Case 1**:  
Capture variation screenshots (4 patterns):  

```
@ComposePreviewOptions(advanceTimeMillis = [100, 3000], themes = [R.theme.xxxDark, R.theme.xxxLight])
```

**Use Case 2**:  
Capture representative patterns (2 configurations):  

```
@ComposePreviewOptions(advanceTimeMillis =100, themes = [R.theme.xxxDark])
@ComposePreviewOptions(advanceTimeMillis =[3000], themes = [R.theme.xxxLight])
```

sergio-sastre: Thanks for showcasing some useful scenarios.
However, my suggestion is that we proceed first to support the scenario without repeatable annotations, and we focus only on the DelayedPreviews.

Let’s provide support for a valuable feature, and then iterate it to make it better 😊

For the other scenario with repeatable annotations, I need to think how to better provide support for it, and it is honestly bad timing: long holidays are approaching together with many other things and I have less time for this than I’d like too till May, likely 😅.

So I propose we focus now only on this DelayPreview, passing an array with intervals at which we want to tale screenshots, and the usecase2 is something we will support later.

what are your thoughts?

takahirom: @sergio-sastre 
I pushed the changes, but there's an error that might be coming from `preview.getAnnotation`.
Could you take a look at this stack trace?

```
java.lang.IllegalArgumentException: Cannot use a ScanResult after it has been closed
	at io.github.classgraph.ScanResult.getClassInfo(ScanResult.java:860)
	at io.github.classgraph.ScanResultObject.getClassInfo(ScanResultObject.java:120)
	at io.github.classgraph.AnnotationInfo.getClassInfo(AnnotationInfo.java:259)
	at io.github.classgraph.AnnotationInfo.convertWrapperArraysToPrimitiveArrays(AnnotationInfo.java:476)
	at io.github.classgraph.ObjectTypedValueWrapper.convertWrapperArraysToPrimitiveArrays(ObjectTypedValueWrapper.java:369)
	at io.github.classgraph.AnnotationParameterValue.convertWrapperArraysToPrimitiveArrays(AnnotationParameterValue.java:167)
	at io.github.classgraph.AnnotationParameterValueList.convertWrapperArraysToPrimitiveArrays(AnnotationParameterValueList.java:116)
	at io.github.classgraph.AnnotationInfo.getParameterValues(AnnotationInfo.java:131)
	at io.github.classgraph.AnnotationInfo.getParameterValues(AnnotationInfo.java:207)
	at com.github.takahirom.roborazzi.AndroidComposePreviewTester.test(RoborazziPreviewScannerSupport.kt:306)
	at com.github.takahirom.roborazzi.RoborazziPreviewParameterizedTests.test(RoborazziPreviewParameterizedTests.kt:38)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at androidx.compose.ui.test.junit4.AndroidComposeTestRule$apply$1$evaluate$1.invoke(AndroidComposeTestRule.android.kt:272)
	
```

takahirom: I think the issue occurs because the [nested annotation](https://github.com/takahirom/roborazzi/pull/633/commits/f21f6ff3dc945183f6249af92503a471db5b0c4f#diff-9249965fa2df22231143fa7040e485d053071c416b3d0472407f66340c249582R7-R15) requires access to ClassGraph after closure (it appears the class graph is closed after `getPreviews()`).

We can work around this by using LongArray instead of annotation arrays, though this may reduce flexibility. Potential solutions include:

1. **Preemptive data storage**: Save all `otherAnnotationsInfo.?...parameterValues` during `getPreviews()` to avoid post-closure access errors. This might introduce performance considerations.

2. **Resource management adjustment**: Remove the `use{}` block in `getPreviews()` to prevent premature closure of scanResult. However, this approach risks resource leaks, as discussed in [ClassGraph issue #233](https://github.com/classgraph/classgraph/issues/233).

3. **Type substitution**: Implement LongArray parameters instead of Annotation types. While this avoids nested data structures efficiently, it may limit future extensibility for complex use cases.

Code comparison:

// Original

```kotlin
@Target(AnnotationTarget.FUNCTION)
annotation class RoboComposePreviewOptions(
  val manualClockOptions: Array<ManualClockOptions> = arrayOf()
)

annotation class ManualClockOptions(
  val advanceTimeMillis: Long = 0L
)
```

// Proposed

```kotlin
@Target(AnnotationTarget.FUNCTION)
annotation class RoboComposePreviewOptions(
  val pausedAdvanceTimeMillis: LongArray = longArrayOf()
)
```

sergio-sastre: @takahirom 
The first idea coming to my mind that could solve it would be to add ManualClockOptions to the `includeAnnotationInfoForAllOf()`, so that annotation info is included in the ScanResult after being closed, and can be resolved by `getAnnotation()`

Could you give it a try? If it does not work, I can take a further look at it :)

takahirom: Thanks for your suggestion. Unfortunately, [it](https://github.com/takahirom/roborazzi/pull/655) doesn't work and produces the same error. The error might be caused by parameterValues, as AnnotationInfoList appears unrelated to `includeAnnotationInfoForAllOf()`.


```
inline fun <reified T : Annotation> ComposablePreview<*>.getAnnotation(): T? {
    val annotationParams = otherAnnotationsInfo
        ?.filter { it.name == T::class.java.name }
        ?.firstOrNull()?.parameterValues // here
        ?: return null
```

I've attempted the fix here:
https://github.com/takahirom/roborazzi/pull/655

sergio-sastre: Yep, after a second thought I realised that is only for annotated composables.

Btw, ehat is the reason to make MainClockOptions an annotation and not a class? Although I do not think it will make a difference.

I’ll see if I find some time to check how to fix that tomorrow, today I cannot.

in the meanwhile, since it is an extension method, you could copy pasted , import Kotlin reflect and try some fixes, as an idea

takahirom: Thanks. Annotations cannot contain class instance members; they only accept primitive types or other annotations. I'll explore the possibility of retrieving annotations using alternative methods.

takahirom: I believe I was able to implement a workaround.

https://github.com/takahirom/roborazzi/pull/633/commits/63610f7813e53bb7c6d82844f50f0006425be559

This is unrelated, but one area I'd like to investigate is whether we can incorporate users' custom ComposeRule or similar components in this implementation.

takahirom: I have removed the `options()` override from AndroidComposePreviewTester and added ComposeRuleFactory to JUnit4TestLifecycleOptions. While this approach helps address the challenges of integrating custom ComposeRules, it introduces significant breaking changes that users need to be aware of during migration.

takahirom: My main concern is whether the KMP Preview support is currently functioning properly. As we currently lack test cases, I would like to establish comprehensive test cases to verify its functionality.

takahirom: I believe we need to significantly refactor this code. The main issue is that we currently cannot capture multiple screenshots using different manualClockOptions because captureScreenshot closes the ActivityScenario. 
I suggest creating separate tests for each screenshot, which would also require introducing a breaking change to the previews() function.

takahirom: I've changed it and will review my code. 👀 
https://github.com/takahirom/roborazzi/pull/633/commits/1d3cade29ad11d0044667f14f7c1601716203381

takahirom:
This is going to be a big release for users using Compose Preview Support. I'm thinking releasing this as 1.44.0-**alpha01** to gather feedback.

---
## Breaking changes for [Experimental Compose Preview Support](https://takahirom.github.io/roborazzi/preview-support.html) users

### Key interface changes
1. `ComposePreviewTester` now handles `TestParameter` instead of raw `ComposablePreview`
2. New `@RoboComposePreviewOptions` annotation enables time-based captures
3. `roborazzi-annotations` module added as new dependency

### Migration steps
1. **Update custom testers**:
```kotlin
// Before
class CustomTester : ComposePreviewTester<AndroidPreviewInfo> {
  override fun previews(): ComposablePreview<AndroidPreviewInfo>  = AndroidComposablePreviewScanner().scanPackageTrees(...)
  override fun test(preview: ComposablePreview<AndroidPreviewInfo>)
}

// After
class CustomTester : ComposePreviewTester<AndroidPreviewJUnit4TestParameter> {
  override fun testParameters(): AndroidPreviewJUnit4TestParameter = AndroidComposablePreviewScanner().scanPackageTrees(...)
  override fun test(preview: AndroidPreviewJUnit4TestParameter)
}
```

2. **Add dependency**:
```gradle
implementation("io.github.takahirom.roborazzi:roborazzi-annotations:")
```

3. **Use new annotation** for time-based captures:
```kotlin
@RoboComposePreviewOptions(
  manualClockOptions = [ManualClockOptions(advanceTimeMillis = 500L)]
)
@Preview
@Composable
fun DelayedPreview() {
  // Content with animations/timed state changes
}
```

### Sample implementations
- [CustomPreviewTester](https://github.com/takahirom/roborazzi/pull/633/files#diff-7d8c7957b1d5f1a8d9e8c4f0e7b7f8e2f5a1b3c) handling Android previews  
- [MultiplatformPreviewTester](https://github.com/takahirom/roborazzi/pull/633/files#diff-3b8d9c7a2d3a8e1b9f9c7d8e5a0b6c7f) for KMP previews

These changes enable annotation-driven screenshot control while maintaining backward compatibility for most preview configurations.

PR #642, opened 2025-01-25
Print collection content in ComposePrintToString to stabilize output in Dump screenshots.

siarhei-luskanau:
Dump screenshot has node properties. Sometime Roborazzi generates new screenshots (or fails verification) due the node print the `androidx.compose.ui.semantics.CollectionInfo` reference value instead of content. This Pull Request add the output sorting to avoid the issue.


takahirom:
Thank you for ensuring Roborazzi's determinism!

PR #644, opened 2025-02-03
Update formatting of left column in reports

mannodermaus:
This is a proposal to adjust the layout of Roborazzi reports a little, addressing some quirks I have found when working with it. Left is before, right is after!

### General updates

I adjusted the default width of the report container to account for larger screens. Instead of materialize's default values (90%, max 1280px), it is now 95% with a maximum width of 1440px. Furthermore, the ratio between the left and right columns is now 40-60, a little different to the 30-70 from before. I also added a little fix for contextData: Before this PR, if any contextData has an empty value, then _no_ context data would be shown at all. This is now changed so that only these empty context data entries are filtered out instead.

### Show relative path to output directory above each file name

This accounts for use cases where a non-default directory is used to store images. The folder is a little smaller than the file name, too.

Screenshot 2025-02-03 at 16 40 08

### Fix display of `contextData` and make it easier to digest

First of all, fix the missing line break. Furthermore, use a little sub-table to display context data key-value pairs more neatly. Add a cute badge thingy to make this data stand out a bit more.

Screenshot 2025-02-03 at 16 40 35

### Fix display of `aiAssertionResults`

Another missing line break fix. Use a badge to separate these results from the main file name. Add some tweaks to color and size to achieve hierarchy of information.

|||
|---|---|
|Screenshot 2025-02-03 at 16 42 05|Screenshot 2025-02-03 at 16 38 10|

### DIFF
--- a/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
+++ b/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
@@ -33,28 +33,10 @@ sealed interface CaptureResult {
     }
 
   @InternalRoborazziApi
-  fun reportText(): String {
-    return buildString {
-      append(reportFile.name)
-      if (contextData.isNotEmpty() && contextData.all {
-          it.value.toString() != "null" && it.value.toString().isNotEmpty()
-        }) {
-        appendLine("contextData:$contextData")
-      }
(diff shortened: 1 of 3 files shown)


eddnav:
I think a unit test for this default formatting of the file name would be nice. This line-break issue broke our tests, but not really, it just can't find the old files because of the name change.

takahirom (approved):
Thank you for your contribution! It looks excellent.

PR #654, opened 2025-02-12
Fix for HTML result will break when context data key has special char…

unni-dm:
When you have special characters in context data (like space), the final result html file will break. 

Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Document': '#Galaxy%20S22%20Ultra' is not a valid selector.

Fix is to remove special characters from id. 

| Before | After |
| ------------- | ------------- |
| Screenshot 2025-02-12 at 13 55 30  |  Screenshot 2025-02-12 at 13 52 36| 
| Screenshot 2025-02-12 at 13 55 45 |  |


takahirom:
Thanks. I'll merge this and release as `1.41.1`

PR #697, opened 2025-06-13
Add dim color to dialog

takahirom:


include-build/roborazzi-core/src/androidMain/kotlin/com/github/takahirom/roborazzi/capture.kt
```diff
@@ -48,6 +50,11 @@ sealed interface RoboComponent {
       rootsOrderByDepth.forEach { root ->
         val layoutParams = root.windowLayoutParams.get()
         val decorView = root.decorView
+        if ((layoutParams.flags and WindowManager.LayoutParams.FLAG_DIM_BEHIND) !== 0) {
+          val alpha = (layoutParams.dimAmount * 255).toInt()
+          val color = Color.argb(alpha, 0, 0, 0)
+          decorView.setBackgroundColor(color)
```
jeppeman, line 56: Looks like we can't use `decorView` here as it's only wrapping the size of the dialog body. 

I noticed that [this](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/layoutlib/bridge/src/android/view/WindowManagerImpl.java;l=172?q=windowmanagerim) was in layoutlib, not the actual WindowManagerImpl running on Android; so they're just mimicking the behaviour of a [Dimmer](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/services/core/java/com/android/server/wm/Dimmer.java). So I think it's reasonable to do something similar here, I tried the following and it produced the desired results:

```kotlin
if ((layoutParams.flags and WindowManager.LayoutParams.FLAG_DIM_BEHIND) !== 0) {
  val alpha = (layoutParams.dimAmount * 255).toInt()
  val color = Color.argb(alpha, 0, 0, 0)
  canvas.drawRect(Rect(0, 0, width, height), Paint().apply { this.color = color })
}
```

takahirom: It seems to be working correctly now. I had initially tried this, but I misunderstood and thought the semi-transparent color was filling the area with a solid color. However, I now realize that it is being drawn with the proper transparency.

PR #715, opened 2025-08-08
Minor tweak to documentation for Groovy DSL

smugleafdev:
I found the existing documentation for using Groovy DSL instead of Kotlin DSL confusing. This is one minor change applied to two different locations, with clearer wording and the second example included since these conversions can be non-obvious.


takahirom (approved):

Issue #703, opened 2025-06-25
Can you add changeThreshold to RoboComposePreviewOptions or the generateComposePreviewRobolectricTests extension?


pavel163:
Currently, it's not possible to set a threshold for the generated test for a Compose preview.
It would be great to have the ability to do this either via the RoboComposePreviewOptions annotation, a new annotation, or through the extension (which would be less flexible).

takahirom:
You can implement your own `ComposePreviewTester` and include your desired options by following the documentation here: https://takahirom.github.io/roborazzi/preview-support.html#generate-compose-preview-screenshot-tests. Nevertheless, we are always open to API suggestions and would be happy to consider them.

https://github.com/takahirom/roborazzi/blob/main/roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt#L234C37-L234C57

takahirom:
This is just a preliminary idea, but I'm considering providing an interceptor for users. We'll likely need some time and feedback to finalize the design.


```kotlin
class YourAndroidComposePreviewTester : ComposePreviewTester<AndroidPreviewJUnit4TestParameter> by AndroidComposePreviewTester(
  interceptors = listOf(
    object: RoborazziOptionsInterceptor() {
      override roborazziOptionsIntercept(
        chain: RoborazziOptionsInterceptor.Chain,
      ): RoborazziOptions {
        return chain.proceed(chain.roborazziOptions).copy(
          ...
        )
      }
    }
  )
)
```

takahirom:
@pavel163 Thank you for your feedback, we are working on this. If you have any opinion on this implementation, let me know!
https://github.com/takahirom/roborazzi/pull/716

takahirom:
Thank you for your feedback! I have made the default behavior customizable. Regarding RoboComposePreviewOptions, I will discuss it in another issue.

PR #716, opened 2025-08-09
Add Capturer interface for customizing AndroidComposePreviewTester

takahirom:
# What
Add a Capturer interface to AndroidComposePreviewTester to enable customization of screenshot capture behavior.

# Why
Addresses Issue #703 - Users requested the ability to set custom changeThreshold for generated Compose preview tests. The composition-based solution allows users to customize capture behavior while maintaining backward compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1019,7 +1019,9 @@ import com.dropbox.differ.SimpleImageComparator
 import com.github.takahirom.roborazzi.*
 
 @ExperimentalRoborazziApi
-class MyCustomComposePreviewTester : AndroidComposePreviewTester(
+class MyCustomComposePreviewTester(
+  private val defaultCapturer: AndroidComposePreviewTester.Capturer = AndroidComposePreviewTester.DefaultCapturer()
+) : AndroidComposePreviewTester(
   capturer = { parameter ->
     val customOptions = parameter.roborazziOptions.copy(
       compareOptions = parameter.roborazziOptions.compareOptions.copy(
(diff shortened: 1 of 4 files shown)


takahirom:
@sergio-sastre Thanks as always! 🙏
This PR addresses #703 - adds a way to customize RoborazziOptions (like changeThreshold) in generated Compose Preview tests.
Since this builds on ComposablePreviewScanner, would love your feedback on the approach!

Key: Added a `Capturer` interface to AndroidComposePreviewTester for customizing capture behavior while maintaining backward compatibility

sergio-sastre:
Sure! I’ll take a look at this this weekend whenever I find time 😊

sergio-sastre (commented):
@takahirom 
I think it might be more comfortable for the user to just set a general threshold via the roborazzi plugin (which would be overriden by the annotation value for those previews with it), so they would not have to deal with the tester.

Moreover, one could make it even more flexible by extending the  `RoboComposePreviewOptions` to add a threshold parameter.

In doing so, the user could set a different threshold for each preview separately via the annotation. If not threshold set, the user would default to the "general threshold" set in the plugin option.

takahirom:
@sergio-sastre 
Thank you for your feedback! I agree that setting the threshold from the user's perspective would be more comfortable. However, I think that if we add a public API for threshold now, the next requests will probably be for APIs for hshift, file name, scale, and so on. If we keep adding more and more public APIs, it may become unmanageable. For now, I think introducing this change will help reduce the pain for users, so I would like to proceed with it as is.

PR #721, opened 2025-08-13
Fix preview size retention bug

### LINKED ISSUE #719: Incorrect preview snapshot sizes
The recent 1.46.x and 1.47.x releases have broken something in the way snapshot sizes are determined. It seems to be related to the `heightDp` and `widthDp` attributes in the Preview annotations.

### Context 

We are using Roborazzi together with the ComposablePreviewScanner to automatically generate snapshots for our previews. Basically we use the scanner to get a list of all the previews and then use a big for loop to capture and assert all the preview snapshots.

### Problem

Some of our components have previews where we fix the height of the preview to check the UI in very small or very tall screens. This was working fine with Roborazzi release verion 1.45.0, but in the latest releases these previews also change the size of other snapshots.

```kotlin
@Preview
@Composable
fun MyComponent_Preview1() {
     Text("default size")
}

// Smaller preview results in a smaller snapshot.
@Preview(heightDp = 600)
@Composable
fun MyComponent_Preview2() {
    Text("small size")
}

// This preview should have a snapshot with default size but it is only 600dp high!
@Preview
@Composable
fun MyComponent_Preview3() {
    Text("default size again")
}
```

So it feels like that if one preview fixes the height at 600dp then the following previews also gets 600dp height, even if they have no `heightDp` attribute.

### How to reproduce

I created a minimal reproduction project [here](https://github.com/smuldr/roborazzi-preview-sizing).

takahirom:
## Summary
Fixes a bug where Preview annotation size attributes (heightDp/widthDp) incorrectly persist between captures when using ComposablePreviewScanner.

## Problem
- When one preview has `heightDp=600`, subsequent previews without explicit size specifications also get rendered at 600dp height instead of default size
- Root cause: `RuntimeEnvironment.setQualifiers()` modifies Robolectric's global display configuration which persists between captures

## Solution
Save and restore Robolectric qualifiers around each capture operation:
- Save current qualifiers before applying configuration
- Restore original qualifiers in finally block after capture
- Keep `RoborazziComposeOptions` immutable (no mutable fields)

## Test
Added `PreviewSizeRetentionBugTest` that reproduces the bug and verifies the fix works by checking actual image dimensions.

Fixes #719

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents preview/screenshot configuration from leaking between captures, improving reliability and consistency across tests and runs.

* **Tests**
  * Added test suites that reproduce and validate preview size retention issues, ensuring the fix is covered.

* **Chores**
  * Increased CI screenshot comparison job timeout to reduce flaky failures on slower runs.


takahirom:
@coderabbitai review

PR #722, opened 2025-08-14
Require ComposablePreviewScanner 0.7.0+ with version validation

takahirom:
# What
Updated Roborazzi to require ComposablePreviewScanner version 0.7.0 or higher, adding automatic version validation in the Gradle plugin to prevent compatibility issues.

# Why
ComposablePreviewScanner 0.7.0 introduced breaking changes with AndroidPreviewScreenshotIdBuilder package restructuring. Without version enforcement, users with older versions would experience build failures with unclear error messages.

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
@@ -242,6 +242,7 @@ fun verifyGenerateComposePreviewRobolectricTests(
       return@afterEvaluate
     }
     verifyLibraryDependencies(project)
+    verifyComposablePreviewScannerVersion(project)
     verifyAndroidConfig(androidExtension, logger)
   }
 }
@@ -326,4 +327,42 @@ private fun verifyLibraryDependencies(
     "io.github.sergio-sastre.ComposablePreviewScanner:android",
   )
   requiredLibraries.forEach { allDependencies.checkExists(it) }
+}
+
+private fun verifyComposablePreviewScannerVersion(
+  project: Project
+) {
+  val dependencies = project.configurations.flatMap { it.dependencies }
+  val composablePreviewScannerDependency = dependencies.find { 
+    it.group == "io.github.sergio-sastre.ComposablePreviewScanner" && it.name == "android" 
+  }
+  
+  if (composablePreviewScannerDependency != null) {
+    val version = composablePreviewScannerDependency.version
+    if (version != null && isVersionLessThan(version, "0.7.0")) {
+      error(
+        "Roborazzi: ComposablePreviewScanner version 0.7.0 or higher is required. " +
+        "Current version: $version. " +
+        "Please update your ComposablePreviewScanner dependency to version 0.7.0 or higher."
+      )
+    }
+  }
+}
+
+private fun isVersionLessThan(currentVersion: String, requiredVersion: String): Boolean {
+  val current = parseVersion(currentVersion)
+  val required = parseVersion(requiredVersion)
+  
+  for (i in 0 until maxOf(current.size, required.size)) {
+    val currentPart = current.getOrNull(i) ?: 0
+    val requiredPart = required.getOrNull(i) ?: 0
+    
+    if (currentPart < requiredPart) return true
+    if (currentPart > requiredPart) return false
+  }
+  return false
+}
+
+private fun parseVersion(version: String): List<Int> {
+  return version.split(".").mapNotNull { it.toIntOrNull() }
 }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -48,7 +48,7 @@ webjar-materialize = "1.0.0"
 webjars-locator-lite = "0.0.6"
 webpImageio = "0.3.3"
 
-composable-preview-scanner = "0.6.1"
+composable-preview-scanner = "0.7.0"
 
 [libraries]
 accessibility-test-framework = { module = "com.google.android.apps.common.testing.accessibility.framework:accessibility-test-framework", version.ref = "accessibilityTestFramework" }
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/PreviewGenerateTest.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/PreviewGenerateTest.kt
@@ -71,6 +71,17 @@ class GeneratePreviewTestTest {
       }
     }
   }
+
+  @Test
+  fun whenUsingOldComposablePreviewScannerVersionShouldBeError() {
+    RoborazziGradleRootProject(testProjectDir).previewModule.apply {
+      buildGradle.composablePreviewScannerVersion = "0.6.1"
+
+      record(BuildType.BuildAndFail) {
+        assert(output.contains("Roborazzi: ComposablePreviewScanner version 0.7.0 or higher is required"))
+      }
+    }
+  }
 }
 
 class PreviewModule(
@@ -87,6 +98,7 @@ class PreviewModule(
     private val PATH = moduleName + "/build.gradle.kts"
     var isKmp = false
     var includePreviewScannerSupportDependenciy = true
+    var composablePreviewScannerVersion = "0.7.0"
     fun write() {
       val file =
         projectFolder.root.resolve(PATH)
@@ -168,7 +180,7 @@ class PreviewModule(
                           $previewScannerSupportDependency
                           implementation(libs.junit)
                           implementation(libs.robolectric)
-                          implementation(libs.composable.preview.scanner)
+                          implementation("io.github.sergio-sastre.ComposablePreviewScanner:android:$composablePreviewScannerVersion")
                           implementation(libs.androidx.compose.ui.test.junit4)
                       }
                   }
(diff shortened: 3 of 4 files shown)


takahirom:
@sergio-sastre 
Could you take a look at this PR? It updates ComposablePreviewScanner to version 0.7.0 and adds version validation to show clear error messages when users have incompatible older versions. The changes help users upgrade smoothly by providing guidance when version requirements aren't met.

sergio-sastre:
> I found this to be a difficult problem to avoid because I didn't know that moving APIs to a superclass is source-code compatible, but not bytecode compatible.

Yes, I did not suspect it'd not be bytecode compatible since it is "inherited" as you said.

However, Google created the [metalava](https://android.googlesource.com/platform/tools/metalava/+/HEAD/COMPATIBILITY.md) tool to detect such incompatibility API changes, and we're using a [gradle metalava plugin](https://github.com/tylerbwong/metalava-gradle) in my current project indeed 😅.

It generates a .txt file containing the signatures of the public api and has a a validate task to verify API compatibility at all levels.

Integrating this in Github actions should help detect these changes at least, and that is also my idea...

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
```diff
@@ -327,3 +330,82 @@ private fun verifyLibraryDependencies(
   )
   requiredLibraries.forEach { allDependencies.checkExists(it) }
 }
+
+private fun verifyComposablePreviewScannerVersion(
+  project: Project
```
sergio-sastre, line 335: Oh I see!
I thought that the import in the generated code was the problem, but it is in the roborazzi module that provides support for composablePreviewScanner right?

I think that would explain why it is source code compatible but not at bytecode level...

In that case I think it is only fixable by enforcing the user to use 0.7.0+ when using Roborazzi. So this looks fine :)

include-build/roborazzi-gradle-plugin/src/test/java/io/github/takahirom/roborazzi/VersionComparisonTest.kt
```diff
@@ -0,0 +1,57 @@
+package io.github.takahirom.roborazzi
+
+import org.junit.Test
+import org.junit.Assert.assertTrue
+import org.junit.Assert.assertFalse
+
+class VersionComparisonTest {
+
+    @Test
```
sergio-sastre, line 9: mega minor:
I think a parameterized test makes more sense here, since Junit uses hard asserts by default: once the first assert fails, the test does not execute more code.

That is the reason why soft-assertions where created.
https://www.baeldung.com/java-assertj-soft-assertions

And with Parameterized tests, every assert would run in a separate test, which it is even better than soft-assertions when possible, in my opinion.
This should not be a blocker though, feel free to change it or not :)

sample-generate-preview-tests-multiplatform/src/androidUnitTest/kotlin/com/github/takahirom/preview/tests/MultiplatformPreviewTester.kt
```diff
@@ -8,8 +8,8 @@ import com.github.takahirom.roborazzi.*
 import com.github.takahirom.roborazzi.ComposePreviewTester.TestParameter.JUnit4TestParameter
 import org.junit.rules.RuleChain
 import org.junit.rules.TestWatcher
-import sergio.sastre.composable.preview.scanner.jvm.common.CommonComposablePreviewScanner
-import sergio.sastre.composable.preview.scanner.jvm.common.CommonPreviewInfo
+import sergio.sastre.composable.preview.scanner.common.CommonComposablePreviewScanner
+import sergio.sastre.composable.preview.scanner.common.CommonPreviewInfo
 
 @OptIn(ExperimentalRoborazziApi::class)
```
sergio-sastre, line 14: This test is still not accounting for all the new preview parameters that were added to `@Preview`s in common?
```kotlin
override fun test(testParameter: JUnit4TestParameter<CommonPreviewInfo>) {
    val preview = testParameter.preview
    val previewInfo = preview.previewInfo
    val screenshotName = CommonPreviewScreenshotIdBuilder(preview).build()
    val filePath = "$DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH/$screenshotName.png"
    captureRoboImage(
      filePath = filePath,
      roborazziComposeOptions = RoborazziComposeOptions {
        size(previewInfo.widthDp, previewInfo.heightDp)
        locale(previewInfo.locale)
        background(
          showBackground = previewInfo.showBackground,
          backgroundColor = previewInfo.backgroundColor
       )
      }
    )
}
```

And also add a preview with a given width, height and background, for instance, extra annotate one of the previews with this:
```kotlin
@Preview(widthDp = 400, heightDp = 400, showBackground = true, backgroundColor = 0xFF0000FF)
```

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
takahirom: I suspect the issue is that the bytecode for `roborazzi-compose-preview-scanner-support` invokes `AndroidPreviewScreenshotIdBuilder.ignoreClassName`. I'm guessing a `NoSuchMethodError` occurs because that method doesn't exist.

takahirom:
> However, Google created the [metalava](https://android.googlesource.com/platform/tools/metalava/+/HEAD/COMPATIBILITY.md) tool to detect such kind of incompatibility API changes, and we're using a [gradle metalava plugin](https://github.com/tylerbwong/metalava-gradle) in my current project at work indeed 😅.

I wasn't introducing them intentionally to avoid slowing us down. But I think we can add and commit api.txt automatically so we don't have to fail the CI. I think we can just track the API changes.

I was thinking of using this library. The decision is tough because metalava is better, but it might not support KMP (klib) well. It seems that Google is using both tools.
https://github.com/Kotlin/binary-compatibility-validator

sample-generate-preview-tests-multiplatform/src/androidUnitTest/kotlin/com/github/takahirom/preview/tests/MultiplatformPreviewTester.kt
takahirom: Thanks. I think we still can't use preview parameters in this PR because of these errors.
https://github.com/takahirom/roborazzi/pull/722#issuecomment-3190450496

include-build/roborazzi-gradle-plugin/src/test/java/io/github/takahirom/roborazzi/VersionComparisonTest.kt
takahirom: Ah, I see how it works now. It wraps and collects every error to create a single AssertionError. Thanks for clarifying.

Anyway, I've addressed your suggestion in this commit: 6a3c1a5

sergio-sastre:
> > However, Google created the [metalava](https://android.googlesource.com/platform/tools/metalava/+/HEAD/COMPATIBILITY.md) tool to detect such kind of incompatibility API changes, and we're using a [gradle metalava plugin](https://github.com/tylerbwong/metalava-gradle) in my current project at work indeed 😅.
> 
> I wasn't introducing them intentionally to avoid slowing us down. But I think we can add and commit api.txt automatically so we don't have to fail the CI. I think we can just track the API changes.👍
> 
> I was thinking of using this library. The decision is tough because metalava is better, but it might not support KMP (klib) well. It seems that Google is using both tools. https://github.com/Kotlin/binary-compatibility-validator

Oh, I menat I should inteoduce it in ComposablePreviewScanner, this was definitely my bad 😅. Sprry for the confusion

sample-generate-preview-tests-multiplatform/src/androidUnitTest/kotlin/com/github/takahirom/preview/tests/MultiplatformPreviewTester.kt
sergio-sastre: > Thanks. I think we still can't use preview parameters in this PR because of these errors. [#722 (comment)](https://github.com/takahirom/roborazzi/pull/722#issuecomment-3190450496)

is there any reason not to update AGP?
Once updated, if I understood the errors correctly, we can fix the rest of errors by updating only the :sample-generate-preview-tests-multiplatform compileSdk to 35 right?
However, I am unsure if updating AGP might have any undesired consequence 🤔, so I understand your concern

takahirom: According to this compatibility guide, we would need to update to Kotlin 2.1, but it feels a bit early to adopt it. However, I think it might be acceptable to only update the AGP.
https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-compatibility-guide.html#version-compatibility

To avoid compatibility error, I need to add the following flag:
kotlin.mpp.androidGradlePluginCompatibility.nowarn=true

takahirom: Thanks. I've tried to use Compose Multiplatform 1.9.0-beta03 with Preview. We also use Kotlin Native, and it seems that it doesn't work with Kotlin 2.0.21.
https://github.com/takahirom/roborazzi/pull/726#issuecomment-3191211222

Issue #751, opened 2025-10-24
Enable test sharding in the Gradle plugin


sergio-sastre:
I've been thinking for a while about this and I believe it would be a great addition to Roborazzi.

So, there is a wait to parallelise JVM tests by using
```kotlin
testOptions {
   unitTests.all {
       it.maxParallelForks = 5 // your desired value here!
   }
}
```

However, Gradle can only parallelise tests by classes. The thing is, we have ONLY ONE CLASS that takes as parameter all previews.

Therefore, they cannot be parallelised. The idea would be that the plugin can generate several classes, each taking an (almost) equal number of Previews as parameter.

My idea is as follows:
- by default, the plugin default number of shards is the same as the maxParalleForks value.
That would be something like this in the `ComposablePreviewRoborazziPlugin`:
```kotlin
project.afterEvaluate {
            // Default numOfShards to Gradle Test.maxParallelForks (users can still override via extension)
            val tests = project.tasks.withType(org.gradle.api.tasks.testing.Test::class.java)
            val maxForks = tests.findByName("test")?.maxParallelForks
                ?: tests.maxOfOrNull { it.maxParallelForks } ?: 1
            extension.numOfShards.convention(maxForks)

            if (extension.enable.get()) {
                setupGenerateComposablePreviewRoborazziTestsTask(project, extension)
            }
        }
```
- The plugin also accepts a numOfShards parameter for granular control. Here we would have to implement "error handling" to make the user aware that maxParalleForks is necessary if he has not set it.

For passing the previews, we could split them by index, sth similar to this in the `GenerateComposablePreviewRoborazziTestTask`:
```
private fun generateTestFileContent(
        packageName: String,
        className: String,
        packagesExpr: String,
        includePrivatePreviewsExpr: Boolean,
        numShards: Int,
        includeHeader: Boolean
    ): String {
      
      // this are the cached previews, out of the classes but with lazy
     val cachedPreviews = 
      "private val cachedPreviews: List<ComposablePreview<AndroidPreviewInfo>> by lazy {
                AndroidComposablePreviewScanner()
                    .scanPackageTrees($packagesExpr)
                    ${if (includePrivatePreviewsExpr) ".includePrivatePreviews()" else ""}
                    .getPreviews()
            }"
     // here we generate 1 test class for each shard,
     repeat(numShards) { shardIndex ->
        val valuesExpr = 
            "cachedPreviews.filterIndexed { index, _ -> index % $numShards == $shardIndex }"
         
[...]

PR #761, opened 2025-11-28
Add test sharding support for preview tests

takahirom:
# What
Adds test sharding support to the Gradle plugin for parallel execution of Compose Preview tests.

When `numOfShards > 1`, the plugin generates multiple test classes (RoborazziPreviewParameterizedTests0, Tests1, etc.) that can be executed in parallel by Gradle's maxParallelForks.

# Why
Addresses #751 - Compose Preview tests currently cannot be parallelized because Gradle can only parallelize tests by class, and the plugin generates only one parameterized test class containing all previews.

This enables users to leverage maxParallelForks for faster test execution:
- Auto-detects `maxParallelForks` value from test configuration
- Users can override with explicit `numOfShards` setting
- Uses modulo-based distribution for even test distribution across shards
- Maintains backward compatibility (numOfShards=1 generates single class)

## Summary by CodeRabbit

* **New Features**
  * Configurable test-class sharding for preview/Robolectric tests via a new generatedTestClassCount option; auto-detection from maxParallelForks when unset.

* **Tests**
  * Added integration tests covering single- and multi-shard generation and parallel execution scenarios.

* **Documentation**
  * Updated README and preview docs with examples for generatedTestClassCount and guidance on aligning with parallel forks.

* **Chores**
  * Ignored /temp in .gitignore.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
+++ b/include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
@@ -69,21 +69,14 @@ open class GenerateComposePreviewRobolectricTestsExtension @Inject constructor(o
     .convention(false)
 
   /**
-   * The number of test shards to generate for parallel execution.
+   * The number of test classes to generate.
    * By default, this is automatically set to match the test task's maxParallelForks value.
-   * Set this explicitly to override the automatic detection or to disable sharding (numOfShards = 1).
+   * Set this to match maxParallelForks for parallel test execution.
    *
-   * When numOfShards = 1, generates a single test class (backward compatible).
-   * When numOfShards > 1, generates multiple test classes (RoborazziPreviewParameterizedTests0, Tests1, etc.)
-   * Each shard uses modulo-based distribution: shard i contains tests where (testIndex % numOfShards == i)
-   *
-   * Example: With 4 shards and 10 tests:
-   * - Shard 0: tests 0, 4, 8
-   * - Shard 1: tests 1, 5, 9
-   * - Shard 2: tests 2, 6
-   * - Shard 3: tests 3, 7
+   * When generatedTestClassCount = 1, generates a single test class.
+   * When generatedTestClassCount > 1, generates multiple test classes (RoborazziPreviewParameterizedTests0, Tests1, etc.)
    */
-  val numOfShards: Property<Int> = objects.property(Int::class.java)
+  val generatedTestClassCount: Property<Int> = objects.property(Int::class.java)
 
 }
 
@@ -127,11 +120,11 @@ private fun setupGenerateComposePreviewRobolectricTestsTask(
       "See https://github.com/sergio-sastre/ComposablePreviewScanner?tab=readme-ov-file#how-to-use for more information."
   }
 
-  // Auto-detect numOfShards from maxParallelForks if not explicitly set
+  // Auto-detect generatedTestClassCount from maxParallelForks if not explicitly set
   project.afterEvaluate {
-    if (!extension.numOfShards.isPresent) {
+    if (!extension.generatedTestClassCount.isPresent) {
       val maxForks = testTaskProvider.mapNotNull { it.maxParallelForks }.maxOrNull() ?: 1
-      extension.numOfShards.convention(maxForks)
+      extension.generatedTestClassCount.convention(maxForks)
     }
   }
 
@@ -205,7 +198,7 @@ private fun setupGenerateComposePreviewRobolectricTestsTask(
     it.includePrivatePreviews.set(extension.includePrivatePreviews)
     it.testerQualifiedClassName.set(testerQualifiedClassName)
     it.robolectricConfig.set(robolectricConfig)
-    it.numOfShards.set(extension.numOfShards)
+    it.generatedTestClassCount.set(extension.generatedTestClassCount)
   }
   // We need to use sources.java here; otherwise, the generate task will not be executed.
   // https://stackoverflow.com/a/76870110/4339442
@@ -236,7 +229,7 @@ abstract class GenerateComposePreviewRobolectricTestsTask : DefaultTask() {
   abstract val robolectricConfig: MapProperty<String, String>
 
   @get:Input
-  abstract val numOfShards: Property<Int>
+  abstract val generatedTestClassCount: Property<Int>
 
   @TaskAction
   fun generateTests() {
@@ -245,10 +238,10 @@ abstract class GenerateComposePreviewRobolectricTestsTask : DefaultTask() {
 
     val packagesExpr = scanPackageTrees.get().joinToString(", ") { "\"$it\"" }
     val includePrivatePreviewsExpr = includePrivatePreviews.get()
-    val numShards = numOfShards.get()
+    val testClassCount = generatedTestClassCount.get()
 
-    require(numShards >= 1) {
-      "numOfShards must be >= 1, but was $numShards"
+    require(testClassCount >= 1) {
+      "generatedTestClassCount must be >= 1, but was $testClassCount"
     }
 
     val generatedClassFQDN = "com.github.takahirom.roborazzi.RoborazziPreviewParameterizedTests"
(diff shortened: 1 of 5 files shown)


takahirom:
@sergio-sastre 
Thank you for suggesting this. I've created a pull request for test sharding. Could you take a look?

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
```diff
@@ -209,24 +235,82 @@ abstract class GenerateComposePreviewRobolectricTestsTask : DefaultTask() {
...
+
+    // Delete old generated test files to avoid conflicts when changing numOfShards
+    directory.listFiles()?.filter { it.extension == "kt" }?.forEach { it.delete() }
     val robolectricConfigString =
       "@Config(" + robolectricConfig.get().entries.joinToString(", ") { (key, value) ->
         "$key = $value"
       } + ")"
     val testerQualifiedClassNameString = testerQualifiedClassName.get()
+
+    if (numShards == 1) {
```
sergio-sastre: Minor: I'd also handle the edge case `numOfShards <= 0` and throw the corresponding exception

sergio-sastre:
> @sergio-sastre 
> 
> Thanks. I believe we are facing three distinct challenges.
> 
> 
> 
> 1. Forgetting to set maxParallelForks
> 
> I’m still undecided. The optimal number of forks hinges on the hardware users (and CI) possess, and CI environments sometimes run in a single-process mode. If we elevate the omission to a hard error, we will impose cumbersome work-arounds; if we rely solely on a warning, many users will simply ignore it. Another possibility is to raise the error but provide an escape hatch—e.g., acceptProcessAndSettingDifference—though I concede that feels somewhat clumsy. At present I cannot devise a more elegant solution.
> 
> 
> 
> 2. Parallelisation is not working
> 
> I verified whether tests run in separate processes after setting maxParallelForks.
> 
> 
> 
> 
> 
> ```kotlin
> 
> @Test
> 
> fun test() {
> 
>     val pid = try {
> 
>         val processHandleClass = Class.forName("java.lang.ProcessHandle")
> 
>         val currentMethod = processHandleClass.getMethod("current")
> 
>         val pidMethod = processHandleClass.getMethod("pid")
> 
>         val handle = currentMethod.invoke(null)
> 
>         pidMethod.invoke(handle) as Long
> 
>     } catch (e: Exception) { -1L }
> 
>     println("=== $className.test(): PID=$$pid, param=$$testParameter ===")
> 
> ```
> 
> 
> 
> ```
> 
> === RoborazziPreviewParameterizedTests3.test(): PID=53590,
> 
>     param=JUnit4TestParameter(preview=com.github.takahirom.preview.tests.PreviewsKt_PreviewWithProperties2) ===
> 
> === RoborazziPreviewParameterizedTests0.test(): PID=53593, param=JUnit4TestParameter(preview=com.github.takahirom.preview.tests.PreviewsKt_PreviewNormal)
> 
> …
> 
> ```
> 
> 
> 
> Have you set maxParallelForks?
> 
> 
> 
> ```kotlin
> 
> unitTests {
> 
>     all {
> 
>         it.maxParallelForks = 4 // ←
> 
> ```
> 
> 
> 
> 3. Duplicate scanning
> 
> This strikes me as a thorny issue, because we would have to implement some locking mechanism that is both tricky to test and hard to render robust. Accordingly, I propose we defer it to a later stage.

Duplicate scanning per test class

I have to recheck, because I am unsure if it is happening. But this concerns me a bit, since scanning is the most expensive operation, can take up to 5 seconds.
Moreover, the Robolectric test Runner already duplicates the scanner, which I've already reported here

https://github.com/robolectric/robolectric/issues/10605

So if the scanning happens for each class, the test paralellisation could never bring any speed benefit, but rather the opposite:
If you set 4 shards, and scanning takes around 3 seconds... this might mean 24 seconds spent on scanning instead of 6.

takahirom:
It seems that with 4 forked JVM processes, the scan runs 8 times.
I believe the issue is that we have two class loaders: one for calculating parameters and one for execution. If the class loader differs, the class itself—and therefore its companion object—is considered different, and we are unable to cache the parameters. That is why the scan is performed twice within each process.

```
        === RoborazziPreviewParameterizedTests0 SCAN START: PID=79103, ClassLoader=1984697014 ===
         === RoborazziPreviewParameterizedTests1 SCAN START: PID=79105, ClassLoader=1984697014 ===
         === RoborazziPreviewParameterizedTests2 SCAN START: PID=79104, ClassLoader=1984697014 ===
         === RoborazziPreviewParameterizedTests3 SCAN START: PID=79106, ClassLoader=1984697014 ===
         === RoborazziPreviewParameterizedTests3 SCAN START: PID=79106, ClassLoader=1982876486 ===
         === RoborazziPreviewParameterizedTests1 SCAN START: PID=79105, ClassLoader=1442077586 ===
         === RoborazziPreviewParameterizedTests2 SCAN START: PID=79104, ClassLoader=1442077586 ===
         === RoborazziPreviewParameterizedTests0 SCAN START: PID=79103, ClassLoader=1442077586 ===
```

takahirom:
We can obtain testParameters from the parentClassLoader; nevertheless, we cannot treat them at runtime as `List<ComposePreviewTester.TestParameter<*>>`, because `ComposePreviewTester.TestParameter` is nominally the same class yet, owing to the distinct class loader, is effectively a different type, so the cast fails.

takahirom:
I found that ClassGraph's ScanResult has a toJSON function. 👀 
We could use System.setProperty to cache it..?

sergio-sastre:
@takahirom 
Just an update. I've tried by dumping the scan result to a json and although the time seems better when many files are scanned, the scanning is still taking longer than expected. That is because what is dumped is the files, not the previews only, therefore it is still performing the scanning.

I've also realized that my example with Paparazzi also scans 1 per class(instead of 2 because it does not use the Robolectric test runner).

The other option I'll try next would be:
1. Gradle task that generates a file and store the Previews scanned in memory in a singleton.
2. Execute the tests, which access that singleton
3. Delete the file with the singleton

I think this would work for Paparazzi, but might not work for Roborazzi because of the Sandboxmode from Robolectric.

Another option would be to make the  ComposablePreviews serializable and store them in a json (only the previews!) so when scanning that json file, it just contains what has to be scanned instead of everything.

Anyway, I am open to more ideas 😊

sergio-sastre:
@takahirom 

I am finding it pretty challenging, since the ComposablePreviews are not serialisable, and I am also unsure whether that would be the way to go.

I think, I will first check the time winning when using ScanResultDumper to dump the ScanResult to a json.

To be honest, I am testing with very few previews (order 30-50) and I think sharing makes sense when the number is 10 times bigger than that (300-500)

So I'll try to make a benchmark to see what are the times with
- 1 fork
- 2 forks

both considering scanning from packages vs. scanning from json. 

I believe I'll need some days to explore this

sergio-sastre:
Just an update.

I am currently preparing ComposablePreviewScanner 0.8.0, which I'll release in the next 1-3 days, and then I'll look deeply into this. 

Sorry for delaying this a little bit more 🙏

sergio-sastre:
@takahirom 
Hi! Thanks for your patience!
I got some time to do some initial benchmarks with around 1000 Previews and I can confirm that without saving in a file there are significant time wins 🎉

So the average times I had on my Mac i7 are as follows for 
- 1000 `@Preview`s
- 16 packages, 
- 5 executions

1 shard -> 1 min 9 sec
2 shards -> 48 sec
3 shards -> 52 sec
4 shards -> 1 min 2 sec

I've also observed that, although it scans as many times as shards, it happens in parallel (what actually makes sense), so it is not as bad as I thought it could be.

Therefore I believe the current approach is already pretty good for the first version :)

sergio-sastre:
> @sergio-sastre Sorry for being late, and thank you for checking it out! I'm still considering the issue where we forget to set maxParallelForks.
> 
> * If we don't provide any errors or warnings, people will forget to set maxParallelForks and will be confused about why tests can't run in parallel.
> * If we force people to set maxParallelForks by throwing an exception, they might need to change the setting in both local and CI environments, which could be cumbersome.
> 
> What I'm thinking is simply renaming numOfShards to generatedTestClassCount to reduce confusion, since whether to shard or not is ultimately a decision for the test framework to make.

Yes, I like that solution, it represents the reality more accurately and it is pragmatic.
The use case for it can be added in the doc and README.md

Issue #783, opened 2026-01-16
Add excludeIfAnnotatedWithAnyOf property to roborazzi gradle plugin


alecarnevale:
Hi, I'd like to setup roborazzi gradle plugin to generate screenshot test from Composable previews using `generateComposePreviewRobolectricTests`, but I need to instruct roborazzi gradle plugin to ignore some preview annotated with a custom annotation.

The underlying `ComposablePreviewScanner` allows such configuration with a [AndroidComposablePreviewScanner. excludeIfAnnotatedWithAnyOf()](https://github.com/sergio-sastre/ComposablePreviewScanner?tab=readme-ov-file#api).

Would it be possible to add support for such a use case to the roborazzi gradle plugin?

as6o:
I'd guess that introducing an annotation to turn off roborazzi for a @Preview is the primary reason developers have to implement a custom tester at all (that, and configuring compare options).  Anything that can be done to simplify these use cases for developers would be greatly appreciated (well, I'd greatly appreciate it!)

Personally, I'm now upgrading my roborazzi deps and am struggling with the new ComposePreviewTester APIs.  I think I have it working but there wasn't a good "exclude annotation" example that I could find using the new API and copying/overriding `AndroidComposePreviewTester.testParameters()` is disconcerting with the "FIXME" in the impl comments.  Introducing an ignore annotation so that a ComposePreviewTester doesn't have to be written at all would remove these issues and concerns and, I think, make the library easier to use for a lot of people.

I don't know a lot about the ComposablePreviewScanner library, sergio-sastre is obviously the expert, but it is unfortunate if a limitation in a dependency prevents this library from being easier to use.  Sounds like it is not possible to call `excludeIfAnnotatedWithAnyOf` and `includeIfAnnotatedWithAnyOf` at the same time because of the _possibility_ that some annotations _may_ conflict?  I'd think checking for actual conflicts (or introducing a precedence) would be better than a broader limitation.  But, like I said, I'm not as familiar with that library so I may be misunderstanding the discussion here.

At the least, is there a good example of the recommended way to introduce an "ignore" annotation using the newer API?  Maybe even with compare configuration as well?

@sergio-sastre, @takahirom - thanks for your efforts in these libraries, they are great tools for Android developers.

takahirom:
Thanks for the feedback! I hadn't thought about using TDD for this. So, what about an API like this?

The default filter will be `AnnotationFilter.Exclude("RoboComposePreviewIgnore")`.

[...]

PR #818, opened 2026-04-11
Feature/issue #785

sergio-sastre:
#783 #785 
A draft PR to support Include and Exclude annotations for Composable Preview Screenshot Test Generation

## Summary by CodeRabbit

* **New Features**
  * Annotation-based filtering for compose preview test generation.
  * New annotations to explicitly include or exclude previews in test runs.
  * New configurable annotationFilter option for generated test tasks.
  * Sample project updated to demonstrate annotation-based include/exclude usage.

### DIFF
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
@@ -17,8 +17,8 @@ sealed class AnnotationFilter : Serializable {
   }
 
   companion object Filter {
-    val RoboPreviewExclude = Exclude("com.github.takahirom.roborazzi.annotations.filter.RoboPreviewExclude")
-    val RoboPreviewInclude = Include("com.github.takahirom.roborazzi.annotations.filter.RoboPreviewInclude")
+    val RoboPreviewExclude = Exclude("com.github.takahirom.roborazzi.annotations.RoboPreviewExclude")
+    val RoboPreviewInclude = Include("com.github.takahirom.roborazzi.annotations.RoboPreviewInclude")
   }
 
   /**
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
@@ -17,8 +17,8 @@ sealed class AnnotationFilter : Serializable {
   }
 
   companion object Filter {
-    val ExcludeFromRoborazzi = Exclude("com.github.takahirom.roborazzi.annotations.filter.ExcludeFromRoborazzi")
-    val IncludeInRoborazzi = Include("com.github.takahirom.roborazzi.annotations.filter.IncludeInRoborazzi")
+    val RoboPreviewExclude = Exclude("com.github.takahirom.roborazzi.annotations.filter.RoboPreviewExclude")
+    val RoboPreviewInclude = Include("com.github.takahirom.roborazzi.annotations.filter.RoboPreviewInclude")
   }
 
   /**
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
@@ -8,7 +8,8 @@ import java.io.Serializable
 @ExperimentalRoborazziApi
 sealed class AnnotationFilter : Serializable {
   /**
-   * Exclude only previews with the specified annotations.
+   * Exclude only previews with the specified annotations, passing their absolute Path as String.
+   * If the absolute path is a nested class, pass the JVM binary name using `$`, * e.g. `com.example.Outer$Inner`.
    */
   @ConsistentCopyVisibility
   data class Exclude private constructor(val annotations: List<String>) : AnnotationFilter() {
@@ -16,12 +17,13 @@ sealed class AnnotationFilter : Serializable {
   }
 
   companion object Filter {
-    val ExcludeRoborazzi = Exclude("com.github.takahirom.roborazzi.annotations.filter.ExcludeFromRoborazzi")
-    val IncludeRoborazzi = Include("com.github.takahirom.roborazzi.annotations.filter.IncludeInRoborazzi")
+    val ExcludeFromRoborazzi = Exclude("com.github.takahirom.roborazzi.annotations.filter.ExcludeFromRoborazzi")
+    val IncludeInRoborazzi = Include("com.github.takahirom.roborazzi.annotations.filter.IncludeInRoborazzi")
   }
 
   /**
-   * Include only previews with the specified annotations.
+   * Include only previews with the specified annotations, passing their absolute Path as String.
+   * If the absolute path is a nested class, pass the JVM binary name using `$`, * e.g. `com.example.Outer$Inner`.
    */
   @ConsistentCopyVisibility
   data class Include private constructor(val annotations: List<String>) : AnnotationFilter() {
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
@@ -1,0 +1,25 @@
+package com.github.takahirom.roborazzi
+
+import java.io.Serializable
+
+/**
+ * Filter for composable previews by annotation.
+ */
+@ExperimentalRoborazziApi
+sealed class AnnotationFilter : Serializable {
+  /**
+   * Exclude previews with the specified annotations.
+   */
+  @ConsistentCopyVisibility
+  data class Exclude private constructor(val annotations: List<String>) : AnnotationFilter() {
+    constructor(vararg annotations: String) : this(annotations.toList())
+  }
+
+  /**
+   * Include only previews with the specified annotations.
+   */
+  @ConsistentCopyVisibility
+  data class Include private constructor(val annotations: List<String>) : AnnotationFilter() {
+    constructor(vararg annotations: String) : this(annotations.toList())
+  }
+}
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
@@ -8,11 +8,16 @@ import java.io.Serializable
 @ExperimentalRoborazziApi
 sealed class AnnotationFilter : Serializable {
   /**
-   * Exclude previews with the specified annotations.
+   * Exclude only previews with the specified annotations.
    */
   @ConsistentCopyVisibility
   data class Exclude private constructor(val annotations: List<String>) : AnnotationFilter() {
     constructor(vararg annotations: String) : this(annotations.toList())
+  }
+
+  companion object Filter {
+    val ExcludeRoborazzi = Exclude("com.github.takahirom.roborazzi.annotations.filter.ExcludeFromRoborazzi")
+    val IncludeRoborazzi = Include("com.github.takahirom.roborazzi.annotations.filter.IncludeInRoborazzi")
   }
 
   /**
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
@@ -8,8 +8,10 @@ import java.io.Serializable
 @ExperimentalRoborazziApi
 sealed class AnnotationFilter : Serializable {
   /**
-   * Exclude only previews with the specified annotations, passing their absolute Path as String.
-   * If the absolute path is a nested class, pass the JVM binary name using `$`, * e.g. `com.example.Outer$Inner`.
+   * Exclude previews annotated with any of the given annotations.
+   *
+   * @param annotations Fully qualified annotation class names. For nested classes, use the
+   * JVM binary name with `$`, e.g. `com.example.Outer$Inner`.
    */
   @ConsistentCopyVisibility
   data class Exclude private constructor(val annotations: List<String>) : AnnotationFilter() {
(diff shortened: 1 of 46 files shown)


include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Filter for composable previews by annotation.
+ */
+@ExperimentalRoborazziApi
+sealed class AnnotationFilter : Serializable {
+  /**
+   * Exclude previews with the specified annotations.
+   */
+  @ConsistentCopyVisibility
+  data class Exclude private constructor(val annotations: List<String>) : AnnotationFilter() {
```
sergio-sastre, line 17: @takahirom 
Gemini told me that it was better to call a list to avoid gradle caching problems with arrays. Therefore it is calling a list under the hood. I honestly believe it is better to provide only the varargs version, since it is simpler to consume as API from the user's perspective.

Since the Main constructor is private, the @ConsistentCopyVisibility is necessary to compile (Kotlin 2.0.0 requirement).

Feel free to adapt this, it is just to showcase what I had in mind

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
```diff
@@ -67,6 +70,12 @@ open class GenerateComposePreviewRobolectricTestsExtension @Inject constructor(o
    * When generatedTestClassCount > 1, generates multiple test classes (RoborazziPreviewParameterizedTests0, Tests1, etc.)
    */
   val generatedTestClassCount: Property<Int> = objects.property(Int::class.java)
+
+  /**
+   * Filter for composable previews by annotation.
+   */
+  @ExperimentalRoborazziApi
```
sergio-sastre, line 80: I think we could skip the ExperimentalRoborazziApi... but up to you :)

include-build/roborazzi-gradle-plugin/build.gradle
```diff
@@ -14,7 +14,7 @@ dependencies {
   compileOnly gradleApi()
   compileOnly libs.android.tools.build.gradle
   compileOnly libs.kotlin.gradle.plugin
-  implementation(project(':roborazzi-core')) {
+  api(project(':roborazzi-core')) {
```
sergio-sastre, line 17: It was necessary to change from implementation to api to import AnnotationFilter in the :sample-generate-preview-tests gradle file, otherwise it was not visible.

if you think there is a better way to do it, feel free.

sample-generate-preview-tests/build.gradle.kts
```diff
@@ -6,10 +12,13 @@ plugins {
   id("io.github.takahirom.roborazzi")
 }
 
+// Execute ./gradlew :sample-generate-preview-tests:recordRoborazziDebug --rerun-tasks
```
sergio-sastre, line 14: I've tested manually and works.

I am wondering if you have any special requirements in how to "automate" to run it in the pipeline and ensure no regression bugs.

The first idea coming to my mind, is that one could pass the annotationFilter option as gradle parameter and have different jobs that. Sth like:
./gradlew :sample-generate-preview-tests:recordRoborazziDebug -Pinclude="com.github.takahirom.preview.tests.Include"

But maybe you do not want to run them in the pipeline or you have other better ideas.

include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AnnotationFilter.kt
```diff
@@ -0,0 +1,25 @@
+package com.github.takahirom.roborazzi
+
+import java.io.Serializable
+
+/**
+ * Filter for composable previews by annotation.
+ */
+@ExperimentalRoborazziApi
+sealed class AnnotationFilter : Serializable {
```
sergio-sastre, line 9: Here we could also add some default that includes/excludes Previews with an existing Roborazzi annotation to simplify it.

For instance in roborazzi-annotations:
```kotlin
annotation class ExcludeRoborazzi

annotation class IncludeRoborazzi
```

and then in this Annotation filter
```kotlin
companion object RoborazziAnnotation {
    val Exclude = Exclude("path.to.roborazzi.annotation.Exclude")
    val Include = Include("path.to.roborazzi.annotation.Include")
 }
```

and the "default" could be used like
```kotlin
import com.github.takahirom.roborazzi.AnnotationFilter.RoborazziAnnotation

roborazzi {
  generateComposePreviewRobolectricTests {
    enable = true
    packages = listOf("com.github.takahirom.preview.tests")
    annotationFilter = RoborazziAnnotation.Exclude
  }
}
```

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
```diff
@@ -67,6 +70,12 @@ open class GenerateComposePreviewRobolectricTestsExtension @Inject constructor(o
...
+  val annotationFilter: Property<AnnotationFilter> = objects.property(AnnotationFilter::class.java)
```
takahirom, line 81: @coderabbitai
Could you read here and let me know if this is ok with configuration cache.
https://docs.gradle.org/current/userguide/configuration_cache_requirements.html

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -393,6 +416,21 @@ class AndroidComposePreviewTester(
...
+    return map { annotationClassName ->
+      try {
+        val clazz = Class.forName(annotationClassName)
+        if (!clazz.isAnnotation) {
+          throw IllegalArgumentException("The class $annotationClassName is not an annotation")
+        }
+        @Suppress("UNCHECKED_CAST")
+        clazz as Class<out Annotation>
+      } catch (e: ClassNotFoundException) {
+        throw IllegalArgumentException("The annotation class $annotationClassName not found", e)
```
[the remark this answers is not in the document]
sergio-sastre: @coderabbitai
Can you also tell us if Class.forName is enough? What if the annotation has some constructor parameters?

Is there any other safer alternative?

sergio-sastre:
@takahirom feel free to review it when you have time.

I also left some comments with some concerns or suggestions. I'm open to exchange any thought or ideas on it 😊

takahirom:
@sergio-sastre 
Thanks for the PR! Overall this looks great

a few thoughts:

* API shape / implementation → api / @ConsistentCopyVisibility: All fine with me. Agreed that varargs-only is the better user-facing API. ✅ 

* `@ExperimentalRoborazziApi`: I'd prefer to keep it for now since this is a new surface; we can drop it once it stabilizes. ✅ 

* Default Roborazzi annotations (@IncludeRoborazzi / @ExcludeRoborazzi): I like this a lot. It also partly answers the CI question — once we ship these defaults, the sample can annotate a few previews with them and we can run fixed jobs for both the Include and Exclude configurations to verify filtering still works. That won't cover every case but should catch ~80% of regressions without needing -P params or splitting the sample module. 🙏 

```
// roborazzi-annotations annotation class IncludeRoborazzi annotation class ExcludeRoborazzi

// AnnotationFilter.kt 
companion object RoborazziAnnotation {
 val Include = Include("com.github.takahirom.roborazzi.IncludeRoborazzi") 
 val Exclude = Exclude("com.github.takahirom.roborazzi.ExcludeRoborazzi") 
}
```

* Configuration cache: CodeRabbit's analysis looks right to me — Property<AnnotationFilter> with Serializable + reading orNull at execution time is the correct pattern, no concern. ✅ 

* Two items from CodeRabbit I'd like to handle before merge:

- Guard when a custom tester is used. We already throw for includePrivatePreviews when useScanOptionParametersInTester = false and a custom tester is set, because the option silently gets dropped. annotationFilter has the exact same failure mode, so the guard should cover it too:  🙏 

```
if (!useScanOptions && isUsingCustomTester && (includePrivatePreviews || extension.annotationFilter.isPresent)) { 
  throw IllegalArgumentException( "includePrivatePreviews / annotationFilter cannot be set automatically when using a custom tester. ..." ) 
}
```

- Nested annotation class names.  🙏 
`Class.forName("com.example.Outer.Inner")` fails because the JVM binary name uses $ for nested classes. I don't think we need a fallback in code — just a short KDoc note on Include / Exclude so users know to use the $ form:

```
/** * ... * For nested classes, pass the JVM binary name using `$`, * e.g. `com.example.Outer$Inner`. */
```

Thanks again!

sergio-sastre:
> @sergio-sastre 
> 
> Thanks for the PR! Overall this looks great
> 
> 
> 
> a few thoughts:
> 
> 
> 
> * API shape / implementation → api / @ConsistentCopyVisibility: All fine with me. Agreed that varargs-only is the better user-facing API. ✅ 
> 
> 
> 
> * `@ExperimentalRoborazziApi`: I'd prefer to keep it for now since this is a new surface; we can drop it once it stabilizes. ✅ 
> 
> 
> 
> * Default Roborazzi annotations (@IncludeRoborazzi / @ExcludeRoborazzi): I like this a lot. It also partly answers the CI question — once we ship these defaults, the sample can annotate a few previews with them and we can run fixed jobs for both the Include and Exclude configurations to verify filtering still works. That won't cover every case but should catch ~80% of regressions without needing -P params or splitting the sample module. 🙏 
> 
> 
> 
> ```
> 
> // roborazzi-annotations annotation class IncludeRoborazzi annotation class ExcludeRoborazzi
> 
> 
> 
> // AnnotationFilter.kt 
> 
> companion object RoborazziAnnotation {
> 
>  val Include = Include("com.github.takahirom.roborazzi.IncludeRoborazzi") 
> 
>  val Exclude = Exclude("com.github.takahirom.roborazzi.ExcludeRoborazzi") 
> 
> }
> 
> ```
> 
> 
> 
> * Configuration cache: CodeRabbit's analysis looks right to me — Property<AnnotationFilter> with Serializable + reading orNull at execution time is the correct pattern, no concern. ✅ 
> 
> 
> 
> * Two items from CodeRabbit I'd like to handle before merge:
> 
> 
> 
> - Guard when a custom tester is used. We already throw for includePrivatePreviews when useScanOptionParametersInTester = false and a custom tester is set, because the option silently gets dropped. annotationFilter has the exact same failure mode, so the guard should cover it too:  🙏 
> 
> 
> 
> ```
> 
> if (!useScanOptions && isUsingCustomTester && (includePrivatePreviews || extension.annotationFilter.isPresent)) { 
> 
>   throw IllegalArgumentException( "includePrivatePreviews / annotationFilter cannot be set automatically when using a custom tester. ..." ) 
> 
> }
> 
> ```
> 
> 
> 
> - Nested annotation class names.  🙏 
> 
> `Class.forName("com.example.Outer.Inner")` fails because the JVM binary name uses $ for nested classes. I don't think we need a fallback in code — just a short KDoc note on Include / Exclude so users know to use the $ form:
> 
> 
> 
> ```
> 
> /** * ... * For nested classes, pass the JVM binary name using `$`, * e.g. `com.example.Outer$Inner`. */
> 
> ```
> 
> 
> 
> Thanks again!

Thanks! I'll make the necessary adjustments this weekend, I am afraid I won't find enough time before 🙏

sergio-sastre:
@takahirom 
I've addressed both CodeRabbit issues, so I'm done with this.

I've also added some Roborazzi annotations:
@IncludeInRoborazzi / @ExcludeFromRoborazzi

Feel free to adjust the names.

What is still missing:
1. Adjust CI job to ensure that only Previews with Include or without Exclude annotations are executed
2. Documentation about this feature

I changed the gradle file in :sample-preview-generate-tests to show how it should work and test these changes. Feel free to decide what to do there

If you do not mind, I leave those 2 things to you since you know what your preferences are :)

takahirom:
@coderabbitai Could you check the integration test log?

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/GenerateComposePreviewRobolectricTestsExtension.kt
takahirom, line 81: @sergio-sastre I think we should provide a default annotationFilter because we provide RoboPreviewExclude and RoboPreviewInclude, but we don't use them in our default annotation filter. What do you think?

takahirom: I've added `AnnotationFilter.Filter.RoboPreviewExclude` as a default annotationFilter. If you have any opinion on this, please let me know.
https://github.com/takahirom/roborazzi/pull/818/changes/282014d3d5e9c01264d52ac972477ee488897a12

sergio-sastre: @takahirom 
When having a default, there is an addition call to getAnnotation() for each preview, and the user might not need it if not excluding.

Although I do not think it is expensive, it uses some bits of reflection.

I personally prefer not to have a default one, so the user is forced to be more intentional about it, but I do not have any strong opinion on having exclude as default.

takahirom: Sorry for being late. It seems that getAnnotation is pretty fast. I don't think we have to worry about it.



# `getAnnotation()` Reflection Cost — Benchmark Notes

Context: PR #818 adds a default `annotationFilter` (`RoboPreviewExclude`), which makes
the scanner call `getAnnotation()` / `isAnnotationPresent()` once per Compose preview.
Question raised in review: is the reflection overhead significant?

## Benchmark (JDK 21)

| Scenario | per method | per 1000 previews |
|---|---|---|
| Hot loop (same few `Method`s, millions of reads) | ~1 ns | ~0.001 ms |
| Warm (distinct method, 2nd read — cached) | ~80 ns | ~0.08 ms |
| **Cold (distinct method, 1st access — real scan)** | **~5 µs** | **~5 ms** |

## Key point

A naive hot-loop benchmark reports ~1 ns/call, but that's **misleading**: the JVM caches
annotation data on the `Method`/`Class` after first access. A real scan reads each
**distinct** preview method **once**, so the relevant number is the **cold** cost
(~1–5 µs/preview), which includes first-time constant-pool annotation parsing.

## Conclusion

- Realistic overhead: **~1–5 µs per preview** → ~0.25–1 ms for 50–200 previews, ~5 ms for 1000.
- A single preview's screenshot capture takes **seconds**, so the filter cost is **negligible** (well under 0.1%).
- ✅ Shipping a default `annotationFilter` has **no meaningful performance impact**.

_Note: cold figure is a single pre-JIT pass (slightly inflated). For an exact number,
add timing logs to the actual scan in `RoborazziPreviewScannerSupport`._

takahirom:
@coderabbitai Do you still have any concern?

takahirom:
@coderabbitai Could you elaborate on the CI job updates and documentation?

takahirom:
@coderabbitai Thanks, but a couple of these look based on stale info — could you re-check against the current PR head?

1. **Naming**: the annotations are already `RoboPreviewExclude` / `RoboPreviewInclude` (under `com.github.takahirom.roborazzi.annotations`). `IncludeInRoborazzi` / `ExcludeFromRoborazzi` and the `.filter` subpackage were renamed/removed, so the docs already match the code.
2. **CI coverage**: `sample-generate-preview-tests` *is* exercised in CI — `StoreScreenshot.yml` runs an unqualified `./gradlew recordRoborazziDebug`, which includes that module (default `RoboPreviewExclude` filter). The only real gap is the opt-in `RoboPreviewInclude` mode.

Could you take another look at the current head and confirm?

Issue #820, opened 2026-04-15
Absolute path in gradle system properties breaks the remote cache


Link184:
Looks like a few gradle system properties that are used as test task inputs are breaking the remote cache re-usability.

```
roborazzi.compare.output.dir=/Users/vagrant/git/myApp/shared/design-system/build/outputs/roborazzi
roborazzi.output.dir=src/test/snapshots/images
roborazzi.project.path=/Users/vagrant/git/myApp/shared/design-system
roborazzi.result.dir=build/test-results/roborazzi/debug/results,
```

`roborazzi.compare.output.dir` and `roborazzi.project.path` are absolute paths.

Why absolute path is bad as a gradle task input?
We can't reuse the gradle cache if we download it from CI to my local machine because locally I have another absolute path, so gradle will rerun the task again all the time

takahirom:
@Link184 
Thanks. I think it's better not to have an absolute path. But I think the invalidated task will be the test task, and I think the test task will be used with the --tests filter, right? That will also break the cache, so in theory this does not affect much. Let me know if I'm wrong.

Link184:
@takahirom thank for your quick reply. Only the remote cache is not working, incremental build(with local cache, no remote cache) works well. Locally the cache input absolute paths are the same. If filter is the problem then probably incremental build would not work as well as the remote cache build?

takahirom:
Thanks, that makes sense. My point was that in the usual workflow the cache is produced remotely and then consumed locally, and local runs are often done with `--tests`, so I thought the practical impact might be limited.

That said, I agree this is still a valid remote cache relocatability issue.

Link184:
I defined the filter logic in a gradle task and I run it everywhere, on CI and locally. Intuitively if the filter logic is the same the remote cache must work everywhere, I will double check the filtering topic.
Assuming the filtering is breaking the remote cache... for the library it is anyway a win to have only relative paths and make sure the remote cache will work well at least for the projects that are not using filters.
Or maybe it is not so easy to get rid of these 2 absolute paths?

Link184:
I tested the filtering logic and can't find any remote cache related problems. 
I also tried to fix the absolute path and I found that it breaks the logic of KMP ios/native target. I will look for a solution in the next few days. I will create a PR if I find something

Issue #791, opened 2026-01-23
Roborazzi 1.56.0 caches golden_file_path as absolute, invalid after project move


m0jOrty:
Description:
- Roborazzi writes golden_file_path as an absolute path that includes the project folder.
- After copying/moving the project (e.g., to a different CI workspace), the cached path becomes invalid.
- This impacts `verifyRoborazzi` and makes Gradle’s test build cache unusable after project move. Apparently the `testReleaseUnitTest` task cannot be cached
- Setting PathSensitivity.RELATIVE externally did not help.

Expected behavior:
- Cache should be portable: store golden_file_path as a relative path (e.g., relative to roborazzi.outputDir or project/module root), or resolve paths relative to the current context on read.

Environment:
- Roborazzi version: 1.56.0

Attempts:
We have tried to use the the following without success, this also creates the absolute path:
`roborazzi.record.filePathStrategy=relativePathFromRoborazziContextOutputDirectory`

Link184:
@m0jOrty I think the problem is the inability to use the cache and slow build as result. Something similar we have in paparazzi: https://github.com/cashapp/paparazzi/issues/1874

Is it fixable?

takahirom:
@Link184 
Thanks, I'm still weighing it up. Actually, I was thinking that we build the cache on CI and then reuse it locally. But locally we usually run filtered tests, so the cache would be invalidated. In other words, CI runs the full test suite while locally we run only filtered tests, which means the benefit of sharing the cache might be pretty small. That said, it’s definitely worth fixing if we can, so I’ll implement the patch when I get a chance. If you’ve got a solid use-case where the cache really pays off, let me know.

Link184:
> so I’ll implement the patch when I get a chance.

Thank you for a really quick answer and potential fix 🥇 

> If you’ve got a solid use-case where the cache really pays off, let me know.

A multimodule gradle project where the snapshot tests are automatically discovered by the new roborazzi gradle plugin feature :) Many gradle modules depend on each other so modifying one module will trigger a cascade of running tests... the team is using trunk based development(don't ask why :D) and everyone must make sure that he will not break the CI. In this configuration the test runtime becomes crucial and gradle remote cache as a organic way of test filtering, basically is the best way to reduce the test runtime.

takahirom:
Thanks. So you’re relying on CI-to-CI cache, but the checkout path changes from one workflow to another?

Link184:
[...]

PR #824, opened 2026-04-24
Make roborazzi gradle remote cache friendly

Link184:
Make roborazzi test task not consider the absolute paths in gradle task inputs.
Added a new integration test where we check how cache relocation works.

https://github.com/takahirom/roborazzi/issues/820
https://github.com/takahirom/roborazzi/issues/791

## Summary by CodeRabbit

* **Improvements**
  * Enhanced build-cache reuse for verification tasks with improved path sensitivity handling, resulting in better cache efficiency across different project configurations.

### DIFF
--- a/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
+++ b/include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
@@ -685,11 +685,17 @@ class RoborazziGradleProjectTest {
       val goldenDest = File(secondProjectDir.root, "app/build/outputs/roborazzi")
       goldenSource.copyRecursively(goldenDest, overwrite = true)
 
-      // Second project: verify from a different directory, should hit cache
-      // since all inputs (golden images content, task properties) are identical
       RoborazziGradleRootProject(secondProjectDir).appModule.apply {
+        // Verify from a different directory — should hit cache since inputs are identical
         val output = verify().output
         assertFromCache(output)
+
+        // Confirm that when verification fails in secondProjectDir, the actual/compare
+        // output files are written to secondProjectDir (not testProjectDir)
+        changeScreen()
+        verifyAndFail()
+        checkRecordedFileExists("$screenshotAndName.testCapture_actual.png")
+        checkRecordedFileExists("$screenshotAndName.testCapture_compare.png")
       }
     } finally {
       secondProjectDir.delete()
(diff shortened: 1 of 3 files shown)


include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
```diff
@@ -329,7 +330,7 @@ abstract class RoborazziPlugin : Plugin<Project> {
             }
           test.inputs.files(
             imageInputProvider
-          )
+          ).withPathSensitivity(PathSensitivity.RELATIVE)
```
takahirom, line 333: @coderabbitai
I have a question about `withPathSensitivity(PathSensitivity.RELATIVE)`. This seems to mean that we treat the path as a relative path and make it cacheable. What I'm thinking is that the path is absolute and we can save different paths. Can we check it in tests?

include-build/roborazzi-gradle-plugin/src/integrationTest/java/io/github/takahirom/roborazzi/RoborazziGradleProjectTest.kt
```diff
@@ -658,4 +659,40 @@ class RoborazziGradleProjectTest {
...
+      // Copy golden images to the second project so verify has the same file inputs
+      val goldenSource = File(testProjectDir.root, "app/build/outputs/roborazzi")
+      val goldenDest = File(secondProjectDir.root, "app/build/outputs/roborazzi")
+      goldenSource.copyRecursively(goldenDest, overwrite = true)
+
+      // Second project: verify from a different directory, should hit cache
+      // since all inputs (golden images content, task properties) are identical
+      RoborazziGradleRootProject(secondProjectDir).appModule.apply {
+        val output = verify().output
+        assertFromCache(output)
```
takahirom, line 692: @Link184 
Thank you for your pull request!
Could we also assert that `*_actual.png` is written to the expected path(secondProjectDir)?

Right now this test verifies the cache hit, but it does not confirm that `verifyRoborazziDebug` is running against the expected project directory / output path.

takahirom: @coderabbitai Could you write the example code of that?

Link184: @coderabbitai we should not call `verifyAndFail()` in the first project. If the task is failing gradle will not cache it. Correct me if I am wrong

include-build/roborazzi-gradle-plugin/src/main/java/io/github/takahirom/roborazzi/RoborazziPlugin.kt
Link184: I started with the refactoring of `roborazzi.project.path`, basically I put it under a flag since we use it only to bypass the ios target way of working with paths. I reverted the change after I added `withPathSensitivity(PathSensitivity.RELATIVE)` and the new unit test is passing even without the old flag

takahirom:
Now I think I have good confidence in the tests. Thank you for your contribution!

Issue #821, opened 2026-04-20
Roborazzi fails to generate HTML reports with newer `kotlinx.serialization` versions


mhumeniuk-dot:
## Current dependency

In our build, the Roborazzi Gradle plugin resolves `kotlinx.serialization` to a newer version on the shared Gradle plugin classpath:

```text
io.github.takahirom.roborazzi:roborazzi-gradle-plugin:1.59.0
\--- org.jetbrains.kotlinx:kotlinx-serialization-json-jvm:1.6.3 -> 1.10.0
```

## Example error

Task:

```text
:<module>:finalizeTestRoborazziDebug
```

Error:

```text
Execution failed for task ':<module>:finalizeTestRoborazziDebug'.
> 'void kotlinx.serialization.modules.SerializersModuleCollector.contextual(kotlin.reflect.KClass, kotlinx.serialization.KSerializer)'
```

Relevant stacktrace section:

```text
java.lang.AbstractMethodError:
'void kotlinx.serialization.modules.SerializersModuleCollector.contextual(kotlin.reflect.KClass, kotlinx.serialization.KSerializer)'

at kotlinx.serialization.modules.SerialModuleImpl.dumpTo(SerializersModule.kt:194)
at kotlinx.serialization.json.JsonImpl.validateConfiguration(Json.kt:703)
at kotlinx.serialization.json.JsonImpl.<init>(Json.kt:699)
at kotlinx.serialization.json.JsonKt.Json(Json.kt:314)
at com.github.takahirom.roborazzi.CaptureResults.<clinit>(CaptureResults.kt:161)
```

## Request

Please support a newer major `kotlinx.serialization` version in the Roborazzi Gradle plugin runtime.

It would also be beneficial to ensure compatibility with Kotlin 2.x builds.

takahirom:
@mhumeniuk-dot Thanks. Could you tell me your other serialization versions, like `kotlinx-serialization-core`, to clarify the issue?

takahirom:
What I'm thinking is that your kotlinx-serialization-core is somehow outdated and org.jetbrains.kotlinx:kotlinx-serialization-json-jvm is newer.

PR #825, opened 2026-04-28
Avoid contextual module for contextData serialization

mhumeniuk-dot:
## Summary

This changes Roborazzi capture result serialization to use an explicit serializer for `contextData` instead of relying on a contextual serializer registered in the shared `Json` instance.

## Problem https://github.com/takahirom/roborazzi/issues/821

Roborazzi currently builds `Json` with a `SerializersModule` that registers:

```kotlin
contextual(Any::class, AnySerializer)
```

That works when the serialization runtime on the plugin classpath is fully aligned, but it becomes fragile when a newer `kotlinx-serialization-json` is loaded together with an older `kotlinx-serialization-core` version.

In that setup, `Json` initialization can fail during serializers-module validation with:

```text
java.lang.AbstractMethodError:
'void kotlinx.serialization.modules.SerializersModuleCollector.contextual(kotlin.reflect.KClass, kotlinx.serialization.KSerializer)'
```

## Root Cause

The previous implementation depended on contextual serializer resolution at `Json` construction time.

That meant the shared `Json` instance had to validate a serializers module that called into `SerializersModuleCollector.contextual(...)`, which is exactly the ABI-sensitive path that breaks when older `kotlinx-serialization-core` is present on the plugin classpath.

## What Changed

- Removed the global contextual registration for `AnySerializer`
- Replaced `Map<String, @Contextual Any>` with `Map<String, Any>`
- Added `ContextDataSerializer` and applied it directly to `contextData`
- Kept the `contextData` serialization logic local to the field instead of requiring `SerializersModule` configuration

## Why This Is Safer

The new implementation no longer depends on `SerializersModuleCollector.contextual(...)` for `contextData`.

Instead, `contextData` is serialized explicitly through `ContextDataSerializer`, so `CaptureResult` serialization remains self-contained and does not require special `Json` configuration. That avoids the mixed-runtime failure path while preserving the existing behavior for current supported values.

## Behavior

This does not intend to change the JSON contract for current supported `contextData` values:

- `String`
- `Int`
- `Long`
- `Double`
- `Boolean`

## Validation

Ran:

```bash
./gradlew -p include-build --offline :roborazzi-core:jvmTest --tests "io.github.takahirom.roborazzi.CaptureResultTest"
```

and the focused serialization tests passed.

I also published the patched artifacts to Maven Local and updated the standalone repro project to consume them from `mavenLocal()`.

The repro still leaks `kotlinx-serialization-core:1.7.3` from the plugin classpath, but with the patched Roborazzi artifacts the validation run now succeeds.

## Summary by CodeRabbit

* **Refactor**
  * Improved serialization of capture result metadata with enhanced handling of data type conversion to ensure consistent data storage and retrieval.

### DIFF
--- a/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
+++ b/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/CaptureResult.kt
@@ -22,7 +22,7 @@ sealed interface CaptureResult {
   val compareFile: String?
   val actualFile: String?
   val goldenFile: String?
-  val contextData: Map<String, @Contextual Any>
+  val contextData: Map<String, Any>
 
   @InternalRoborazziApi
   val aiAssertionResultsOrNull: AiAssertionResults?
@@ -33,7 +33,7 @@ sealed interface CaptureResult {
     }
 
   @InternalRoborazziApi
-  val contextDataOrNull: Map<String, @Contextual Any>?
+  val contextDataOrNull: Map<String, Any>?
     get() = contextData
       .filter { it.value.toString() != "null" && it.value.toString().isNotEmpty() }
       .takeIf { it.isNotEmpty() }
@@ -53,7 +53,8 @@ sealed interface CaptureResult {
     @SerialName("timestamp")
     override val timestampNs: Long,
     @SerialName("context_data")
-    override val contextData: Map<String, @Contextual Any>
+    @Serializable(with = ContextDataSerializer::class)
+    override val contextData: Map<String, Any>
   ) : CaptureResult {
     override val type = "recorded"
     override val actualFile: String?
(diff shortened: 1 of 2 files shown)


takahirom:
@mhumeniuk-dot Thanks for the PR!

From the issue description, the failure seems to be caused by a mixed kotlinx.serialization runtime: `kotlinx-serialization-json` is resolved to 1.10.0 while an older `kotlinx-serialization-core` is still present on the Gradle plugin classpath.

This change avoids the specific `SerializersModuleCollector.contextual(...)` path, so it may work around the crash, but I’m not sure it fixes the root cause. If the runtime still contains mismatched kotlinx.serialization artifacts, similar ABI issues may appear elsewhere.

Could you first check whether aligning `kotlinx-serialization-core` and `kotlinx-serialization-json` to the same version fixes the repro? If so, I’d prefer fixing the dependency alignment rather than adding a workaround for one serialization path.

mhumeniuk-dot:
@takahirom  Sorry for the delays. Thanks for looking into this. The root cause was a dependency mismatch, not the code path itself. Even so, I think this change is still valuable, because it makes the library more resilient to dependency graph pollution and version skew, and helps prevent similar issues in consumer projects.

takahirom:
Thanks. Even if we merge this PR, we can't have a reproducing test, and we don't know which method can't be used. I think we can't manage the code without knowing that. So I think fixing the dependencies is a good way to solve this. For example, if we have the wrong dependencies, the error will be shown. So let me close this PR.

PR #852, opened 2026-07-03
Add binary compatibility validator with initial API dumps

takahirom:
# What
Introduce [kotlinx binary-compatibility-validator](https://github.com/Kotlin/binary-compatibility-validator) (0.18.1) to the root build and `include-build`, and commit the initial `.api` dumps for all published modules. Sample modules and `roborazzi-idea-plugin` are excluded, and klib (native) validation is disabled — JVM ABI only.

# Why
Upcoming work to commonize the JVM implementation for iOS/KMP support will move code such as `RoborazziOptions` and `processOutputImageAndReport` from JVM source sets to `commonMain`. With `apiCheck` running as part of `check`, any unintended change to the published JVM API surface now fails CI instead of slipping into a release.

## Summary by CodeRabbit

* **New Features**
  * Added AI-based image assertion support with Gemini and OpenAI-compatible models, plus shared assertion options and result reporting.
  * Introduced post-test accessibility checking strategies and options for Compose and View tests.
  * Expanded Compose capture APIs and introduced Compose preview scanning support with configurable presets and generated test integration.
  * Added new Gradle plugin configuration for compare output and preview test generation.
  * Added additional capture utilities for desktop and richer painting/comparison helpers (including GIF encoding).
* **Chores / Tests**
  * Strengthened CI by running API compatibility checks before tests (including in the include-build).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,12 +30,14 @@ jobs:
         id: test
         run: |
           ./gradlew javaDocReleaseGeneration
+          ./gradlew apiCheck --stacktrace
           ./gradlew test --stacktrace -x testReleaseUnitTest
 
       - name: include build tests
         id: include-build-test
         run: |
           cd include-build
+          ./gradlew apiCheck --stacktrace
           ./gradlew test jvmTest --stacktrace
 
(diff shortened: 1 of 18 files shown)


takahirom:
About the diff
It seems that GitHub might be A/B testing the container image.

PR #856, opened 2026-07-03
Fix new golden misreported as Changed in recording tasks

takahirom:
# What
In `processOutputImageAndReport`, a brand-new golden image produced by a recording task type (`VerifyAndRecord` / `CompareAndRecord`) is now reported as `CaptureResult.Added` instead of `CaptureResult.Changed`.

The existence of the golden file is now captured once (`isGoldenFilePresent`) before any write, and that flag is reused for both selecting the golden canvas and the final Added/Changed decision.

# Why
For recording task types the actual image is saved to the golden file itself (`actualFile = goldenFile`). That save happened *before* the final `if (goldenFile.exists())` check, so the check always saw the file it had just written and reported `Changed` for what was actually a newly-added golden.

Flagged during review of #853 (CodeRabbit); pre-existing on main. If #853 (which moves this logic to `commonMain` as `processOutputImageAndReport.common.kt`) merges first, this fix needs a trivial forward-port to that file.

Added a JVM unit test in `roborazzi-core` that runs the `CompareAndRecord` flow with no pre-existing golden and asserts the result is `Added` (RED before the fix: reported `Changed`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected capture result reporting so a missing golden image is now marked as **Added** (not misclassified after a recording step).
  * Improved golden-image handling by determining golden-file presence once per run for more reliable compare/record behavior.

* **Tests**
  * Added a new test covering the image-processing flow when no golden file exists, verifying the correct **Added** result is emitted.

### DIFF
--- a/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.common.kt
+++ b/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.common.kt
@@ -82,7 +82,10 @@ fun processOutputImageAndReport(
   if (taskType.isVerifying() || taskType.isComparing()) {
     val width = (newRoboCanvas.croppedWidth * resizeScale).toInt()
     val height = (newRoboCanvas.croppedHeight * resizeScale).toInt()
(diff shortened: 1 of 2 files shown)


takahirom:
@coderabbitai review

PR #857, opened 2026-07-03
Throw descriptive error when aiAssertionOptions is not configured

takahirom:
# What
`RoborazziOptions.addedAiAssertions()` (and `addedAiAssertion()`) now throw a descriptive `IllegalStateException` when `compareOptions.aiAssertionOptions` is `null`, instead of a bare `NullPointerException` from the `!!` operator. The message explains that an `aiAssertionModel` must be configured in `CompareOptions` first.

# Why
`addedAiAssertions` did `compareOptions.aiAssertionOptions!!.copy(...)`. Called on options built with the default `compareOptions` (`aiAssertionOptions = null`), it threw an unhelpful NPE with no indication of what was missing.

Flagged during review of #853 (CodeRabbit); pre-existing on main. If #853 (which moves this logic to `commonMain` as `RoborazziOptions.common.kt`) merges first, this fix needs a trivial forward-port to that file.

Added a JVM unit test in `roborazzi-core` asserting the `IllegalStateException` type and message (RED before the fix: bare `NullPointerException`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when AI assertion settings are missing, replacing a vague failure with a clearer message.
  * Preserved existing settings updates while safely adding new assertions.

* **Tests**
  * Added coverage to confirm the app now fails predictably and with a helpful message when AI assertion options are not configured.

### DIFF
--- a/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/RoborazziOptions.common.kt
+++ b/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/RoborazziOptions.common.kt
@@ -246,10 +246,16 @@ data class RoborazziOptions(
 
   @ExperimentalRoborazziApi
   fun addedAiAssertions(vararg assertions: AiAssertionOptions.AiAssertion): RoborazziOptions {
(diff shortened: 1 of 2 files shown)


takahirom:
@coderabbitai review

PR #863, opened 2026-07-04
Prototype: touch-robot gesture recording with captureRoboAnimation

takahirom:
Experimental prototype, **stacked on #864** (`tm/compose-1-8-bump`), which is itself stacked on #862 (`tm/capture-robo-animation`). Review/merge #862 and #864 first.

## What this explores
Can [saket/touch-robot](https://github.com/saket/touch-robot) suspend-based input gestures drive a real Compose UI while the new experimental `captureRoboAnimation` API records it frame by frame under Robolectric?

**Answer: yes**, with one small library change.

## The test (sample-android, test scope only)
```kotlin
val touchRobot = rememberTouchRobot(showTaps = true)
LaunchedEffect(Unit) {
  touchRobot.onRoot().performGesture {
    swipe(start = IntOffset(60, 60), stop = IntOffset(360, 360), duration = 500.milliseconds)
  }
}
// ...
composeTestRule.onNodeWithTag("root").captureRoboAnimation(
  composeRule = composeTestRule,
  filePath = "...touch_robot_swipe.gif",
  animationOptions = RoboAnimationOptions(fps = 10),
) {
  delay(1000) // pump virtual time so the LaunchedEffect gesture progresses
}
```
Content is a draggable box; the swipe drags an inner blue box across the frame.

## Dependency: the Maven artifact
Uses `me.saket.touchrobot:touchrobot-core:0.1.0` from Maven Central as a plain `testImplementation` dependency of the sample.

touch-robot is compiled against Compose 1.8+ (its composables call `Composer.shouldExecute`, introduced in runtime 1.8.0), so it needs the Compose 1.8 bump in #864 beneath this PR. An earlier revision of this branch vendored the touch-robot sources to avoid upgrading the classpath; with #864 in place that is no longer necessary, and the `apiVersion` compiler workaround from that revision is also not needed (touch-robot's transitive kotlin-stdlib 2.3.10 on the test classpath provides the `SpillingKt` helper).

## Required library change
touch-robot's `swipe` calls `delay()` between synthesized MotionEvents on the AndroidUiDispatcher (backed by the main Looper). Under Robolectric's PAUSED looper those delayed messages only run when the Looper's virtual clock advances, and `MainTestClock.advanceTimeBy` does **not** advance it — so the gesture stalled right after entering `performGesture` (0 drag events).

Fix: idle the Robolectric main Looper in lockstep with the Compose clock in the recording loop:
```kotlin
composeRule.mainClock.advanceTimeBy(stepMillis)
idleMainLooperFor(stepMillis) // ShadowLooper.shadowMainLooper().idleFor(stepMillis, MILLISECONDS)
composeRule.waitForIdle()
```
Guarded with `catch (NoClassDefFoundError)` so non-Robolectric use is unaffected. Added to both `RoboAnimationRecorderScope.delay` and `recordUntilSettled`. The existing `ComposeAnimationCaptureTest` still passes.

## Validated (per-frame blue-box centroid, 11 frames @ 10fps)
Before fix: box stuck at `(44,44)` every frame, 0 drag callbacks.
After fix: `(44,44) → (60,60) → (140,140) → (236,236) → (308,308) → (328,328)` then settles. The swipe visibly progresses and returns cleanly.

## Not captured: show-taps overlay
touch-robot's show-taps ring (`Color(0xFF0099FF)`) renders at the host-view root as a sibling of the targeted node. `captureRoboAnimation` captures only the targeted semantics subtree (`onNodeWithTag("root")`), so the overlay is outside it and does not appear in the GIF. Capturing it would require recording the full view root rather than a node subtree.

## Summary by CodeRabbit

* **New Features**
  * Added support for recording touch-driven swipe animations in the Android sample.
  * Included a new gesture-based animation capture example that outputs a swipe GIF.

* **Tests**
  * Added a new Compose/Robolectric test to verify swipe gestures are captured correctly during animation recording.

### DIFF
--- a/sample-android/build.gradle
+++ b/sample-android/build.gradle
@@ -91,4 +91,6 @@ dependencies {
   testImplementation libs.androidx.test.core.ktx
   testImplementation libs.robolectric
   testImplementation(libs.kim)
+  // Prototype: touch-robot gesture recording for captureRoboAnimation
+  testImplementation "me.saket.touchrobot:touchrobot-core:0.1.0"
 }
--- a/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ComposeTouchRobotAnimationTest.kt
+++ b/sample-android/src/test/java/com/github/takahirom/roborazzi/sample/ComposeTouchRobotAnimationTest.kt
@@ -1,0 +1,126 @@
+package com.github.takahirom.roborazzi.sample
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.gestures.detectDragGestures
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.unit.IntOffset
+import androidx.compose.ui.unit.dp
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.compose.runtime.LaunchedEffect
+import com.github.takahirom.roborazzi.ExperimentalRoborazziApi
+import com.github.takahirom.roborazzi.RoboAnimationOptions
+import com.github.takahirom.roborazzi.RobolectricDeviceQualifiers
+import com.github.takahirom.roborazzi.captureRoboAnimation
+import com.github.takahirom.roborazzi.roborazziSystemPropertyOutputDirectory
+import me.saket.touchrobot.rememberTouchRobot
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.Assert.assertTrue
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.GraphicsMode
+import kotlin.math.abs
+import kotlin.math.roundToInt
+import kotlin.time.Duration.Companion.milliseconds
+
+/**
+ * Prototype: drive a Compose gesture with saket/touch-robot and record it in real time with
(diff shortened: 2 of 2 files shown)


takahirom:
CI failed on `WindowCaptureTest` ("You need to use a Theme.AppCompat theme"). Root cause: the `me.saket.touchrobot:touchrobot-core:0.1.0` Maven dependency transitively upgraded the test classpath (Compose 1.7.8 -> 1.8.1, kotlin-stdlib -> 2.3.10), and the newer compose test activity broke the appcompat AlertDialog. Excluding transitives doesn't help either: the artifact's composables are compiled against the Compose 1.8+ runtime and fail with `NoSuchMethodError: Composer.shouldExecute` on 1.7.8.

Switched to vendored sources instead (b7f01790):
- Removed the Maven dependency; vendored touch-robot core's five files into `sample-android/src/test/.../sample/touchrobot/` with Apache 2.0 attribution headers. Only the package name changed — no code adaptations were needed for Compose 1.7.8.
- Set Kotlin `apiVersion = 2.1` for `sample-android`: the repo compiles with Kotlin 2.3 but pins kotlin-stdlib to 2.0.x, so with the default apiVersion the vendored suspend functions referenced the stdlib-2.2+ `SpillingKt` helper and crashed with `NoClassDefFoundError`. (Existing tests never hit this because they have no suspend functions.)

Verified: `WindowCaptureTest` (6/6), `ComposeAnimationCaptureTest` (2/2), and `ComposeTouchRobotAnimationTest` (1/1) all green; the recorded GIF still shows the swipe progressing across frames (blue-box centroid (43,43) -> (328,328) over 11 frames). Test runtime classpath is back to baseline: compose ui-android 1.7.8, kotlin-stdlib 2.0.21, no touchrobot artifact.

takahirom:
Reworked to use the `me.saket.touchrobot:touchrobot-core:0.1.0` Maven artifact instead of vendored sources, now that the base branch (#864) bumps the samples to Compose 1.8.3. The vendored copies and the `apiVersion = 2.1` compiler workaround are gone; the branch is now a single commit with the prototype test and the `idleMainLooperFor` recording fix. All touch-robot/window-capture/animation tests pass and the recorded gif shows the swipe progressing (44,44) → (328,328) across 11 frames.

takahirom:
Addressed review feedback:
- The touch-robot test now hoists the dragged offset out of the composable and asserts, after recording, that the box actually moved a substantial amount on both axes. This catches a stalled gesture that would otherwise still produce a (silently identical-frame) GIF and pass.
- The duplicate ShadowLooper / idleMainLooperFor changes were removed from this branch; they now live in the base captureRoboAnimation PR, so this branch retains only the sample test and the touch-robot dependency.

PR #867, opened 2026-07-05
Fix iOS Grid labels overlapping sections by anchoring above margin

takahirom:
## Problem
In the iOS `UIImageRoboCanvas` Grid comparison output, the section labels ("Reference" / "Diff" / "New") were drawn INSIDE the image sections, overlapping the screenshot content. `drawLabel` was called with `destY = margin` and rendered the label box downward from that point (`oy = destY + ly`), so the box covered the top of each section.

On the JVM (`AwtRoboCanvas.drawStringWithBackgroundRect`) `y = margin` is treated as a BASELINE and the box is drawn upward, so its bottom edge lands exactly at the image top edge.

## Fix
`drawLabel` now interprets its `destY` (renamed `bottomY`) as the box's BOTTOM edge, anchored at `margin`. The box is drawn upward (`topY = bottomY - boxHeight`); rows that fall above `y = 0` are clipped at the canvas top edge, matching how the JVM clips a box whose top lands above the canvas. This keeps the labels in the top margin, above the sections, and pins the bottom edge at `margin` for JVM parity.

## Test
- Added `gridLabelsSitAboveSectionsNotOverThem`: builds a grid compare from two solid-color canvases (grid tiers disabled so labels are the only non-golden pixels) and asserts (1) label pixels exist ABOVE `y = margin` under the label x-range, and (2) the reference section's top-edge pixel retains the golden solid color rather than the label background.
- Updated `gridComparisonCanvasHasMarginsGridAndLabels` to look for the label glyph pixels in the top margin band above the section instead of below it.

Both assertions were verified to fail before the fix and pass after (red-then-green). The regenerated tiny-fixture compare PNG confirms the labels now sit in the top margin rather than over the section content.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grid comparison label placement so labels render above the content area instead of overlapping.
  * Updated label drawing to use a bottom-edge anchor, keeping label backgrounds from affecting section pixels.

* **Tests**
  * Added iOS regression tests to ensure labels appear in the top margin and never overwrite sections.
  * Added JVM tests covering rejection of reserved golden filename suffixes (for compare/actual outputs).

* **Bug Fixes**
  * Golden filename reserved-suffix validation now runs earlier, matching prior observable error ordering.

### DIFF
--- a/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.kt
+++ b/include-build/roborazzi-core/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.kt
@@ -18,6 +18,11 @@ fun processOutputImageAndReport(
   canvasFactoryFromFile: CanvasFactoryFromFile,
   comparisonCanvasFactory: ComparisonCanvasFactory,
 ) {
+  // Validate the golden file name BEFORE building the context data so the
+  // reserved-suffix check happens first, matching the observable ordering of the
+  // former monolithic implementation. The common impl re-checks afterwards,
+  // which is harmless.
+  validateGoldenFileNameOrThrow(goldenFile.absolutePath)
   processOutputImageAndReport(
     newRoboCanvas = newRoboCanvas,
     goldenFilePath = goldenFile.absolutePath,
--- a/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.common.kt
+++ b/include-build/roborazzi-core/src/commonMain/kotlin/com/github/takahirom/roborazzi/processOutputImageAndReport.common.kt
@@ -33,6 +33,25 @@ fun interface ComparisonCanvasFactory {
   ): RoboCanvas
 }
 
+/**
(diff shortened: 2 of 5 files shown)


takahirom:
@coderabbitai review

takahirom:
Follow-up review fixes:

- iOS Grid draw order now matches the JVM: grid lines and labels are drawn first, then the reference/diff/new sections are composited on top (via setPixel, which overwrites), so grid lines no longer cross the screenshot content and only show through in the margins/background. Extended the iOS painter tests to assert section content keeps its exact source color on grid-line coordinates while grid pixels remain in the margins. Visually confirmed the regenerated compare PNG.
- Restored validation-before-contextData ordering in the JVM processOutputImageAndReport facade: the reserved-suffix (_compare/_actual) check now runs before buildContextData(), matching the former monolithic behavior. Extracted a shared internal validateGoldenFileNameOrThrow helper reused by the common impl so there is no duplicated string literal, and added JVM unit tests. Public API unchanged (helper is internal).
- Reworded the drawLabel KDoc to describe row clipping (rows above the canvas are skipped) instead of clamping topY to 0.

Verified: :roborazzi-painter:iosSimulatorArm64Test, :roborazzi-core:jvmTest, and apiCheck all green.

takahirom:
@coderabbitai review

PR #868, opened 2026-07-06
Fix iOS result dir resolution and contextData flag handling

takahirom:
These fixes came out of a pre-release review of the iOS KMP stack, ahead of the 1.66.0 release.

## P1 — iOS result-directory creation resolved differently from the write path
`JsonOutputCaptureResultReporter.init` creates the result dir from the RAW (possibly project-relative) path, while `report()` writes to `roborazziToAbsolutePath(...)`, which resolves relative paths against `roborazzi.project.path`. When the project path differs from the simulator process CWD, the directory was created in the wrong place and the JSON write failed (KotlinxIo does not create parents). The iOS `roborazziMakeDirectories` / `roborazziFileExists` actuals now route through `roborazziToAbsolutePath` first, matching the write path (the resolver already returns absolute inputs unchanged). Added `FileSystemSupportTest` coverage that sets `roborazzi.project.path` to a temp dir != CWD and asserts creation/existence land under the project path.

## P2a — include-build iOS tests were not run in CI
The "include build tests" step only ran `apiCheck` + `test jvmTest`, so `roborazzi-core`'s iosTest (including `FileSystemSupportTest`) never ran. Added `./gradlew :roborazzi-core:iosSimulatorArm64Test` to that step (the job already runs on `macos-latest`).

## P2b — iOS ignored the roborazzi.contextdata flag
On JVM the facade honors `roborazziEnableContextData()` (empty map when false); on iOS the flag was ignored. Extracted a common `applyContextDataPolicy(options)` helper used by both the JVM facade (which still layers its JVM-only class-name default on top) and the iOS call site. iOS with flag false → empty context data; flag true → user-supplied contextData only (no className defaults). Documented the JVM-only default in the iOS feature matrix and regenerated README.

## Summary by CodeRabbit

* **New Features**
  * iOS support now respects user-provided context data while keeping default context details JVM-only.
  * CI now runs additional iOS simulator arm64 coverage during build verification.

* **Bug Fixes**
  * Improved iOS file and directory handling so relative paths resolve consistently from the project location.

* **Documentation**
  * Clarified the iOS support matrix to distinguish user-supplied context data from default context data.

* **Tests**
  * Added coverage for iOS context data handling and path resolution behavior.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -41,6 +41,7 @@ jobs:
           cd include-build
           ./gradlew apiCheck --stacktrace
           ./gradlew test jvmTest --stacktrace
+          ./gradlew :roborazzi-core:iosSimulatorArm64Test --stacktrace
 
       - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         if: ${{ always() }}
--- a/README.md
+++ b/README.md
@@ -1385,7 +1385,7 @@ The currently implemented features for iOS support are as follows:
 | threshold / resultValidator | supported |
 | diffPercentage | supported |
 | resizing image (resizeScale) | supported |
-| context data | supported |
+| context data | supported (user-supplied only; the automatically-added default context data such as the test class name is JVM-only) |
 | custom reporter | supported |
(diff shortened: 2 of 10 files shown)


takahirom:
@coderabbitai review

PR #882, opened 2026-07-16
Bound integration-test JVM memory to prevent OOM-killed executors

takahirom:
# What

- Cap the TestKit daemon's metaspace in the integration-test projects' `gradle.properties` (`-XX:MaxMetaspaceSize=1g`).
- Reduce the integration-test workflow's outer Gradle heap from `-Xmx6g` to `-Xmx3g`, with a comment explaining the constraint.

# Why

`RoborazziGradleProjectTest > canRecordWhenRemoveOutputDirBeforeTests` failed in CI ([run](https://github.com/takahirom/roborazzi/actions/runs/27460169315/job/81172287008)) with the inner Robolectric test executor killed by exit 137 (SIGKILL, i.e. out of memory).

The job runs on `macos-latest` (~7GB RAM) and stacks three JVMs: the outer Gradle allowed up to 6g, the shared TestKit daemon (2g heap but **unbounded metaspace**, which grows monotonically across the ~34 sequential TestKit builds due to AGP/Kotlin plugin classloader churn — the failure happened on build 34, test executor 21), and the Robolectric executors with off-heap native-graphics memory. Under that pressure the OS kills the executor, surfacing as a rare, unreproducible test failure.

Capping metaspace stops the unbounded growth (and turns a future overflow into a diagnosable OOM error instead of a runner-level SIGKILL), and the smaller outer heap restores headroom for the nested JVMs. The outer build only compiles the plugin and drives JUnit, so 3g is ample.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved integration test stability by adjusting memory allocation for nested test processes.
  - Increased available JVM metadata space to help prevent failures during extended test runs.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -7,7 +7,11 @@ on:
   pull_request:
 
 env:
-  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx6g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+  # Keep the outer build's heap modest: this job spawns nested JVMs (the TestKit
+  # daemon and Robolectric test executors) that must fit in the same runner
+  # memory. An oversized outer heap led to the test executor being OOM-killed
+  # (exit 137) on long TestKit runs.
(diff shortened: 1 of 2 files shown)


takahirom:
@coderabbitai review

PR #917, opened 2026-08-19
Skip the pixel comparison when images are already identical

AlexanderGH:
Comparing two canvases walks every pixel of both images and allocates a color for each one. Identical images are the common case: it is what a passing screenshot test compares against its golden, and deciding whether a newly captured frame changed usually amounts to the same question.

Compare the pixels a row at a time first, which answers that with bulk integer comparisons, and only fall back to the ImageComparator when they differ. Rows are addressed through the raster's stride so that the sub image backing a cropped canvas is handled too. Equal pixels always mean the comparator would have reported no differences, so nothing is missed. Images that are not laid out as one packed int per pixel keep going through the comparator as before, as do the ones it treats as equal anyway such as transparent pixels whose color channels differ.

Measured on a 1080x2340 image, warmed up, comparing identical content: 15.3ms through SimpleImageComparator versus 0.37ms comparing the pixels directly.

On an Android app module's suite of 65 Roborazzi screenshot tests this removes the comparison from the profile entirely, with SimpleImageComparator.compare going from 200 of 1584 execution samples to none, and all of Roborazzi and differ from 342 samples to 112. The wall clock effect there was small, around 0.2 to 0.7 seconds out of 19 to 25 seconds, because that suite is not bound by this work; the benefit is the CPU and the per pixel allocations it no longer does.

## Summary by CodeRabbit

* **Performance**
  * Improved image comparison speed for pixel-identical images.
  * Optimized comparisons involving cropped images and supported packed-pixel formats.

* **Bug Fixes**
  * Ensured differences in pixels, image dimensions, unsupported formats, and cropped areas receive full comparison handling.
  * Prevented custom image formats with different color mappings from being incorrectly treated as identical.
  * Correctly ignores pixels outside the selected crop region.

* **Tests**
  * Added coverage for identical, differing, cropped, packed-pixel, and custom-format images.

### DIFF
--- a/roborazzi-painter/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AwtRoboCanvas.kt
+++ b/roborazzi-painter/src/commonJvmMain/kotlin/com/github/takahirom/roborazzi/AwtRoboCanvas.kt
@@ -19,6 +19,8 @@ import java.awt.font.TextLayout
 import java.awt.geom.AffineTransform
 import java.awt.image.AffineTransformOp
 import java.awt.image.BufferedImage
+import java.awt.image.DataBufferInt
+import java.awt.image.SinglePixelPackedSampleModel
 import java.io.File
 
 class AwtRoboCanvas(width: Int, height: Int, filled: Boolean, bufferedImageType: Int) : RoboCanvas {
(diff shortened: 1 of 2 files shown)


takahirom:
Thanks! Verified locally — all new tests pass. Nice that the fast path only kicks in for byte-identical rasters of the same type, so nothing changes for the comparator path. The TYPE_CUSTOM test is a nice touch.

Issue #926, opened 2026-09-14
Set Of Mark numbers are too small in Full Screen screenshots


sergio-sastre:
When Using the Ui Tree dump in a Preview with width = 300 and height = 800, I've realised that the marks (the numbers) are very small, what makes them hard to read.

Here an example

<img width="990" height="2200" alt="Image" src="https://github.com/user-attachments/assets/c19f33c2-b5ab-46f6-9849-2804193c43ff" />

Would it be possible to default size of the number marks proportional to the screen size? Although I imagine that might bring other problems when some components are small...  Maybe an option to change the font size of the mark per test in the TestRule/captureRoboImage() would also be helpful (and additionally as gradle property too maybe?)

sergio-sastre:
I think the simplest would be to keep it as it is but provide an option in the TestRule/captureRoboImage() that allows to set the font size of the numbers in UiTreeDumpMode per test.

For Preview Screenshot Tests, an option in the form of a Roborazzi Annotation would be the most optimal, which would use that TestRule/captureRoboImage() option under the hood. This would also mean, we could still provide custom font size per test.

Just a suggestion though

takahirom:
Thanks for the report! I'm looking into this. The easy fix is to size the mark in dp instead of fixed pixels, and I tried that — but with bigger labels they start overlapping the UI underneath, and overlapping each other when a parent and child share the same corner. So I'm still thinking about how to place them.

sergio-sastre:
I've been thinking about this. 
An option would be to provide (maybe not as default) an alternative rendering option similar to the Accessibility dump, which has arrows connecting the Ui Component with a rectangle containing accessibility info, but would contain the corresponding set of marks number in a big font size instead.

I think it would not look as nice as the current version, which is very direct visually (and love that) but would avoid the problems with overlapping.

I'm just brainstorming 😅

PR #929, opened 2026-09-19
Add renderScale for generated Compose Preview Robolectric tests

takahirom:
Stacked on #932. The prelaunch configuration rule and capture use the same renderScale and original device baseline, including previews with no explicit device.

## Summary

Adds an experimental `renderScale` option to generated Compose Preview Robolectric tests.

```kotlin
roborazzi {
  generateComposePreviewRobolectricTests {
    renderScale = 1.0 / 3
  }
}
```

`renderScale` scales the rendering density before capture while preserving the preview's logical dp dimensions. This changes the rendering surface pixel resolution and can reduce rendering work.

It is independent of `resizeScale`, which resizes the bitmap after capture.

## Notes

* Defaults to `1.0`, preserving the existing behavior.
* The option is a `Double`. In the Groovy DSL write `renderScale = 0.5d`; a bare `0.5` is a `BigDecimal` and fails to convert.
* Supports default, named, and `spec:` preview devices.
* Pixel-based device specs are resolved to dp before applying the scale.
* The resulting dpi is rounded to the nearest integer and clamped to at least `1dpi`.
* Density-qualified resources may resolve differently at the scaled density.

## Custom testers

`ComposePreviewTester.Options.renderScale` is a public constructor property, so `super.options().copy(...)` carries the configured value through. A new public `ComposablePreview<AndroidPreviewInfo>.toRoborazziComposeOptions(renderScale)` overload lets a tester that implements `test()` itself apply the value.

A tester that drops the configured scale fails the generated test with an explanation instead of silently recording an unscaled screenshot, so no opt-in flag is needed for this option.

## ComposablePreviewScanner compatibility

The scale is applied by rewriting the dpi qualifier that `RobolectricDeviceQualifierBuilder` emits rather than by calling `Device.copy()`. `Device.copy()` is not binary compatible across the versions Roborazzi supports (`chinSize` changed from `Int` to `ChinSize` after 0.7.0), which made previews with an explicit `device` throw `NoSuchMethodError` on scanner 0.7.0.

## Testing

Added coverage for:

* scaled configuration at Activity startup, without an additional recreation
* avoiding repeated scaling for previews without an explicit device
* default, downscaled, and upscaled rendering
* logical dp size preservation
* rendered surface and screenshot dimensions
* named and custom device specs
* `fontScale` and `uiMode`
* interaction with `resizeScale`
* invalid values and minimum dpi clamping
* detecting a custom tester that drops the configured scale

Local validation: the scanner-support unit tests, the renderScale plugin validation tests, API checks, and two Gradle TestKit integration tests (recording at `renderScale = 0.5` with the default tester, and the failure a scale-dropping custom tester produces) pass.

## Summary by CodeRabbit

- **New Features**
  - Added experimental `renderScale` support for Compose preview Robolectric tests.
  - Configure density scaling while preserving logical dp dimensions across device and preview configurations.
  - Scale values now use double precision and must be finite and positive.
  - Custom testers must preserve configured scaling and scan options; mismatched or missing scaling is reported.

- **Documentation**
  - Clarified `renderScale` versus post-capture `resizeScale`, including output dimensions and golden-image re-recording.
  - Added Kotlin and Groovy configuration guidance.

- **Tests**
  - Added coverage for validation, density handling, device configurations, dimensions, option preservation, and rendered output.


roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RoborazziPreviewScannerSupport.kt
```diff
@@ -324,6 +343,16 @@ interface ComposePreviewTester<TESTPARAMETER : TestParameter<*>> {
...
+    /** Internal bridge used by generated Preview tests. */
+    @InternalRoborazziApi
+    var renderScale: Float = 1f
+      set(value) {
+        require(value.isFinite() && value > 0f) {
+          "renderScale must be finite and greater than 0, but was $value"
+        }
+        field = value
+      }
+
```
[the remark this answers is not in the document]
takahirom: Fixed in a01ea6fb. `renderScale` is now backed by a validated primary-constructor value, so `Options.copy()` preserves it; the API dump and a regression test were updated.

takahirom:
@coderabbitai review

takahirom:
@coderabbitai full review

roborazzi-compose-preview-scanner-support/src/main/java/com/github/takahirom/roborazzi/RenderScaleVerification.kt
```diff
@@ -0,0 +1,53 @@
...
+ * unscaled density without anything failing. [PreviewRenderScaleOption] marks the scale as
+ * applied, and the generated test checks the mark once the capture has run.
+ */
+@InternalRoborazziApi
+object RenderScaleVerification {
+  private var applied = false
+
+  internal fun markApplied() {
+    applied = true
+  }
```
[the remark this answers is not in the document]
takahirom: Fixed in 1e1609e8. `RenderScaleVerification` now records the effective scale of every capture instead of a single Boolean, and `afterTest()` fails unless every recorded capture used the configured value (an empty record still means the scale never reached the capture). Added unit tests for the mismatched scale and the partial multi-capture cases.

takahirom:
@coderabbitai review

takahirom:
@coderabbitai resume

takahirom:
@coderabbitai review

docs/topics/preview_support.md
```diff
@@ -43,12 +43,32 @@ roborazzi {
...
+    // Experimental: scale render resolution while preserving logical dp size.
+    // renderScale = 1.0 / 3
+
     // Filter previews by annotation. See "Filtering previews by annotation" below.
     annotationFilter = AnnotationFilter.Filter.RoboPreviewInclude
   }
 }
 ```
 
+`renderScale` scales rendering density before capture while preserving logical dp dimensions.
```
takahirom, line 56: @sergio-sastre 
Any opinion on this before I ship it? I'm adding `renderScale` to generated Compose Preview tests: it scales the density before rendering while keeping each preview's logical dp size, so the same layout renders onto fewer pixels. It costs fidelity, but 2-3x faster previews seem worth having as an opt-in (`@ExperimentalRoborazziApi`, default `1.0`).

I'm thinking of releasing tomorrow. Here's the release note I'm planning:

---

# Faster preview screenshots

Two independent changes. The first costs you nothing; the second is a trade-off you opt
into. Upgrading alone makes generated preview tests **1.6x faster per preview**, and
combining it with `renderScale` takes that to **3.1x at `0.5`** and **3.7x at `1/3`**.

## Generated preview tests are 1.6x faster per preview

Generated Compose Preview tests no longer recreate the Activity and Compose configuration
for every preview.

| | per preview | goldens |
|---|---|---|
| 1.74.0 | 153.5 ms | 736 KB |
| 1.75.0 | 96.0 ms | 736 KB (byte-identical) |

Nothing to configure, and your recorded images do not change.

## New: `renderScale` (experimental)

`renderScale` scales rendering density while each preview's logical dp dimensions stay the
same, so the layout is unchanged and only the pixel count drops.

```kotlin
roborazzi {
  generateComposePreviewRobolectricTests {
    enable = true
    packages = listOf("com.example.previews")
    renderScale = 0.5
  }
}
```

| `renderScale` | per preview | vs default scale | vs 1.74.0 | pixels recorded | goldens |
|---|---|---|---|---|---|
| `1.0` (default) | 96.0 ms | — | 1.60x faster | 100% | 736 KB |
| `0.5` | 49.0 ms | 1.96x faster | **3.13x faster** | 25% | 292 KB |
| `1.0 / 3.0` | 41.0 ms | 2.34x faster | **3.74x faster** | 11% | 184 KB |

The `vs 1.74.0` column is what you actually see when you upgrade and set the value in the
same step.

Three things to weigh before lowering it:

- **Your screenshots hold fewer pixels**, so a small visual regression can fall below the
  resolution you record at. Re-record your goldens when you change this value.
- **Anything sized in raw pixels does not scale with it.** `dp` and `sp` follow the
  density and keep their proportions, but a value given in pixels — `drawLine(...,
  strokeWidth = 1f)`, a px offset — stays that many pixels in a smaller image, so it comes
  out relatively thicker and in a different place. Previews built on px measurements are
  not simply a smaller version of the original.
- **Returns fall off quickly.** Going from `0.5` to `1/3` cuts the pixels by more than
  half but buys only 1.2x more speed, because what remains is per-preview setup rather
  than rendering. `0.5` is a reasonable starting point.

This is `@ExperimentalRoborazziApi`: the name and behaviour may still change, and we are
looking for feedback on it. If you try it, tell us what you measured and whether the
trade-off worked for your suite in https://github.com/takahirom/roborazzi/pull/929.

The saving is proportional to the area each preview renders, so a suite of large previews
benefits a lot and a suite of default-size previews barely at all. Measure before
committing to a value.

<details>
<summary>How these numbers were measured</summary>

From this repository's own `sample-generate-preview-tests` module (26 Compose previews).
Its sources and Gradle configuration are byte-identical between tag `1.74.0` and
1.75.0, so every configuration renders exactly the same previews.

- Task: `:sample-generate-preview-tests:recordRoborazziDebug` only — record, no compare
- Output images and test results deleted before every run
- `maxParallelForks = 1` (the Gradle default)
- 2 warm-up runs, then 10 measured runs per configuration, configurations interleaved
- Per-preview figures are the median of 250 samples (25 previews x 10 runs), read from the
  JUnit XML `<testcase time>`, with the **first preview of each run excluded**: it absorbs
  JVM, Robolectric and Compose initialisation (2.9-3.2 s in every configuration, ranges
  overlapping) and would otherwise hide the per-preview cost entirely
- Apple M4 Max / 64 GB / macOS 26.3, JDK 21.0.8, Gradle 9.5.1, Robolectric 4.14.1

Whole-task times, for reference — these include Gradle and JVM start-up and are not
comparable to the per-preview figures above:

| configuration | JUnit suite | Gradle invocation (warm daemon) |
|---|---|---|
| 1.74.0 | 7.76 s | 9.6 s |
| 1.75.0, `renderScale = 1.0` | 6.33 s | 8.2 s |
| 1.75.0, `renderScale = 0.5` | 4.68 s | 6.2 s |
| 1.75.0, `renderScale = 1.0 / 3.0` | 4.10 s | 5.7 s |

</details>

sergio-sastre: > @sergio-sastre 
> Any opinion on this before I ship it? I'm adding `renderScale` to generated Compose Preview tests: it scales the density before rendering while keeping each preview's logical dp size, so the same layout renders onto fewer pixels. It costs fidelity, but 2-3x faster previews seem worth having as an opt-in (`@ExperimentalRoborazziApi`, default `1.0`).
> 
> I'm thinking of releasing tomorrow. Here's the release note I'm planning:
> 
> ---
> 
> # Faster preview screenshots
> 
> Two independent changes. The first costs you nothing; the second is a trade-off you opt
> into. Upgrading alone makes generated preview tests **1.6x faster per preview**, and
> combining it with `renderScale` takes that to **3.1x at `0.5`** and **3.7x at `1/3`**.
> 
> ## Generated preview tests are 1.6x faster per preview
> 
> Generated Compose Preview tests no longer recreate the Activity and Compose configuration
> for every preview.
> 
> | | per preview | goldens |
> |---|---|---|
> | 1.74.0 | 153.5 ms | 736 KB |
> | 1.75.0 | 96.0 ms | 736 KB (byte-identical) |
> 
> Nothing to configure, and your recorded images do not change.
> 
> ## New: `renderScale` (experimental)
> 
> `renderScale` scales rendering density while each preview's logical dp dimensions stay the
> same, so the layout is unchanged and only the pixel count drops.
> 
> ```kotlin
> roborazzi {
>   generateComposePreviewRobolectricTests {
>     enable = true
>     packages = listOf("com.example.previews")
>     renderScale = 0.5
>   }
> }
> ```
> 
> | `renderScale` | per preview | vs default scale | vs 1.74.0 | pixels recorded | goldens |
> |---|---|---|---|---|---|
> | `1.0` (default) | 96.0 ms | — | 1.60x faster | 100% | 736 KB |
> | `0.5` | 49.0 ms | 1.96x faster | **3.13x faster** | 25% | 292 KB |
> | `1.0 / 3.0` | 41.0 ms | 2.34x faster | **3.74x faster** | 11% | 184 KB |
> 
> The `vs 1.74.0` column is what you actually see when you upgrade and set the value in the
> same step.
> 
> Three things to weigh before lowering it:
> 
> - **Your screenshots hold fewer pixels**, so a small visual regression can fall below the
>   resolution you record at. Re-record your goldens when you change this value.
> - **Anything sized in raw pixels does not scale with it.** `dp` and `sp` follow the
>   density and keep their proportions, but a value given in pixels — `drawLine(...,
>   strokeWidth = 1f)`, a px offset — stays that many pixels in a smaller image, so it comes
>   out relatively thicker and in a different place. Previews built on px measurements are
>   not simply a smaller version of the original.
> - **Returns fall off quickly.** Going from `0.5` to `1/3` cuts the pixels by more than
>   half but buys only 1.2x more speed, because what remains is per-preview setup rather
>   than rendering. `0.5` is a reasonable starting point.
> 
> This is `@ExperimentalRoborazziApi`: the name and behaviour may still change, and we are
> looking for feedback on it. If you try it, tell us what you measured and whether the
> trade-off worked for your suite in https://github.com/takahirom/roborazzi/pull/929.
> 
> The saving is proportional to the area each preview renders, so a suite of large previews
> benefits a lot and a suite of default-size previews barely at all. Measure before
> committing to a value.
> 
> <details>
> <summary>How these numbers were measured</summary>
> 
> From this repository's own `sample-generate-preview-tests` module (26 Compose previews).
> Its sources and Gradle configuration are byte-identical between tag `1.74.0` and
> 1.75.0, so every configuration renders exactly the same previews.
> 
> - Task: `:sample-generate-preview-tests:recordRoborazziDebug` only — record, no compare
> - Output images and test results deleted before every run
> - `maxParallelForks = 1` (the Gradle default)
> - 2 warm-up runs, then 10 measured runs per configuration, configurations interleaved
> - Per-preview figures are the median of 250 samples (25 previews x 10 runs), read from the
>   JUnit XML `<testcase time>`, with the **first preview of each run excluded**: it absorbs
>   JVM, Robolectric and Compose initialisation (2.9-3.2 s in every configuration, ranges
>   overlapping) and would otherwise hide the per-preview cost entirely
> - Apple M4 Max / 64 GB / macOS 26.3, JDK 21.0.8, Gradle 9.5.1, Robolectric 4.14.1
> 
> Whole-task times, for reference — these include Gradle and JVM start-up and are not
> comparable to the per-preview figures above:
> 
> | configuration | JUnit suite | Gradle invocation (warm daemon) |
> |---|---|---|
> | 1.74.0 | 7.76 s | 9.6 s |
> | 1.75.0, `renderScale = 1.0` | 6.33 s | 8.2 s |
> | 1.75.0, `renderScale = 0.5` | 4.68 s | 6.2 s |
> | 1.75.0, `renderScale = 1.0 / 3.0` | 4.10 s | 5.7 s |
> 
> </details>
> 

@takahirom If I understood correctly, the main benefit is for Previews with big dimensions (e.g. Tablet size).

I find it pretty good and meaningful for the first iteration, and think that having an option to change the default render scale value per Preview (e.g. via Roborazzi annotation) would be a nice improvement.

Just wondering... How does this affect dump mode (e.g. accessibility and Ui Tree)?

takahirom: Great suggestion — I prototyped it and it works: #934.

```kotlin
@RoboComposePreviewOptions(renderScale = 0.5)
@Preview(widthDp = 400, heightDp = 400)
@Composable
fun MyLargePreview() { ... }
```

That preview renders at half resolution, the Gradle-level value stays at its default, and every other preview is untouched. A preview's own value wins over the Gradle one, which defaults to 1.0.

What convinced me was the per-preview breakdown of the benchmark: at 0.5 the largest previews got 3.2-3.5x, mid-size ones 1.3-2.2x, and the smallest ones nothing measurable. One global value is the wrong shape for that.

I did think about an automatic mode (a pixel budget, so only large previews scale down), but then the scale is invisible in the output. A designer looking at a golden can't tell whether it's the real size, and changing a preview's dp a little would silently re-scale it and churn the golden. So I'd rather keep it explicit. Do you think an automatic mode is worth having anyway?

On dump/uiTree mode: it does follow the density, but the overlay is drawn in absolute pixels (basicSize = 600, textSize = 30F/16F), so the labels keep their size and take a larger share of a smaller image. That's how it already behaves at any resolution, so nothing new here.

sergio-sastre: @takahirom I’d rather keep it explicit too, it is also more intuitive for the user than sth automatic I believe.

Regarding the dump/ui tree mode, sounds good. I’ll give it a try once released, and give you feedback on #926

